PR #21, opened 2015-07-02
Fix bug in readme example for creating person

kapilkale:
Hey guys - as I was integrating I noticed this bug. Naming that key `address_state` makes the create command return a 400 error from your server.


alainmeier:
Oops, great call! Thank you.

kapilkale:
NP

Issue #22, opened 2015-08-21
The Person#question_sets method returns an empty array.


DelmerGA:
One would expect that `Person#question_sets` would return the array of `id`'s for the `question_sets` associated to a `Person`. However, the method returns an empty array.

backus:
I agree it should be fixed.  I'm not sure if the expectation should be that we get an array of `id`s.  Instead, I would thing it would be an array lazily requested `QuestionSet` objects.

connorjacobsen:
Dammit. @DelmerGA do you still have the offending code sample? I thought that was something I fixed previously.

I also agree with @backus that it should return objects rather than `id`s, both for consistency across the library, and because users will expect objects rather than strings.

connorjacobsen:
It appears the issue is with how `Person` objects are initialized. After the `super` call, a new `QuestionSet` collection is instantiated, and given the correct `Person` `id`, but existing `QuestionSets` are not loaded, even though we should already have their `id`s from when the `Person` was fetched.

The question, @backus, is what assumptions do we / can we make about the `Person` object? Should we assume that a given `Person` may have been updated by some other object somewhere else in a user's program since the `QuestionSets` were loaded? Or is that an unlikely scenario?

backus:
How would they without going through the client?

connorjacobsen:
I meant if they somehow had two separate objects in their program that both retrieved the same `Person`, and then they did some operations on one of the objects (i.e., created a new question set), and then called the `question_sets` method on the other object.

backus:
That would be fine.  If they had two blockscore instances, fetched a person with the first, and created a question set for that person with the second, then the expected state for the first would still be that nothing about it changes.  That isn't the recommended use case and, if someone did want that for some reason, they could call the `#refresh` method.

connorjacobsen:
:+1: Then I think the fix would be as simple as updating the `QuestionSet` `Collection` upon `Person` instantiation and / or changing the `question_sets` reader into a proper method that handles the lazy loading.

DelmerGA:
I made the suggestion for `Person#question_sets` returning an array of id's because it currently returns something like that. Assuming we have  some `Person` that already has a few `QuestionSet`s or not we see the following:

```ruby
person = BlockScore::Person.retrieve("some_id")
#=> #<BlockScore::Person: ...>
person.question_sets
#=> []  
person.question_sets.create
#=>#<BlockScore::QuestionSet: ... >
person.question_sets.create
[...]

PR #23, opened 2015-08-23
Fix/person question sets array

backus:
Failing test for issue mentioned in #22.  Good to go @DelmerGA


lib/blockscore/base.rb
```diff
@@ -45,16 +59,14 @@ def self.api_url
     end
 
     def self.endpoint
-      if self == Base
-        fail NotImplementedError, 'Base is an abstract class, not an API resource'
-      end
+      fail NotImplementedError, 'Base is an abstract class, not an API resource' if self == Base
```
backus: We could change the conditional to just `if equal?(Base)` which is cleaner and more strict

lib/blockscore/collection.rb
```diff
@@ -1,18 +1,91 @@
...
+      item = target.new(params.merge(default_params))
+      ctxt = self
+      item.define_singleton_method(:save) do
+        ctxt.parent.save unless ctxt.parent_id
+        send :"#{ctxt.parent_name}_id=", ctxt.parent_id
+        super()
+      end
+      self << item
+      item
+    end
```
backus: Is `ctxt` necessary here?

lib/blockscore/base.rb
DelmerGA: SGTM

lib/blockscore/collection.rb
DelmerGA: when removed I believe the QuestionSet instance tries to call `parent` on itself

lib/blockscore/collection.rb
```diff
@@ -1,18 +1,91 @@
...
+
+    def default_params
+      {
+        :"#{parent_name}_id" => parent_id
+      }
+    end
+
+    def parent_id
+      parent.id if parent.respond_to?(:id)
```
backus: When would parent not respond to `id`?

DelmerGA: The issue is that

```ruby
person = BlockScore::Person.new
person.id 
#=> NoMethodError: undefined method `id' for #<BlockScore::Person:0x3fded1db262c JSON:{
}>
```

because it checks the method_missing in `Base` checks attributes then fails, and I wanted to allow someone to be able to build a `Person`  and work with the collection still.

backus: Hm ok.  What do you think about just defining an explicit id method in the resource base class?  Something simple like

```ruby
def id
  attributes.fetch(:id, nil)
end
```

so that classes like `Collection` don't have to introspect on resource classes like this

DelmerGA: Yeah, I like that modification.

lib/blockscore/collection.rb
backus: Huh ok

lib/blockscore/collection/member.rb
```diff
@@ -0,0 +1,87 @@
...
+      #   question_set = QuestionSet.new
+      #   Member.new(person, question_set).save
+      #
+      # @return return value of instance `#save`
+      #
+      # @api public
+      def save
+        save_parent
+        send(:"#{parent_name}_id=", parent.id)
+        instance.save
```
DelmerGA: Bug: saved instances don't have their `id`'s added to the collections id's. 
Will add failing spec and fix. 
Also I'll look into potentially refactoring the tracking strategy of id's of embedded on the parent and newly created items added to collection.

lib/blockscore/base.rb
```diff
@@ -31,6 +49,7 @@ def save
 
     def save!
       response = self.class.post(self.class.endpoint, attributes)
+      # binding.pry
```
backus: pry

lib/blockscore/base.rb
```diff
@@ -4,19 +4,37 @@ module BlockScore
   class Base
     extend Connection
 
-    attr_reader :attributes
-
-    def initialize(options = {})
+    def initialize(options = {}, &block)
+      @loaded = !(block)
```
backus: I'm not sure I understand what is going on here.  I looked at the inheriting classes and I can't seem to find a block being provided

DelmerGA: the block is provided by `retrieve`

lib/blockscore/base.rb
```diff
@@ -4,19 +4,37 @@ module BlockScore
...
-      "#<#{self.class}:0x#{object_id.to_s(16)} JSON: " + JSON.pretty_generate(attributes)
+      str_attr = "JSON:#{JSON.pretty_generate(attributes)}"
+      "#<#{self.class}:0x#{object_id.to_s(16)} #{str_attr}>"
     end
 
     def refresh
-      r = self.class.retrieve(id)
-      @attributes = r.attributes
+      res = self.class.retrieve(id)
+      @attributes = res.attributes
```
backus: I know you didn't write this part but these two lines could likely just be one, right?

DelmerGA: yeah, that looks good

lib/blockscore/base.rb
```diff
@@ -31,6 +49,7 @@ def save
 
     def save!
```
backus: While looking at this I noticed the other save method has a bare rescue which means it just swallows up ALL errors.  I thought I caught this back when the gem was first released but I guess not.  Would you mind fixing that and double checking this isn't done elsewhere in the lib?

lib/blockscore/base.rb
```diff
@@ -45,16 +64,18 @@ def self.api_url
...
     end
 
+    def persisted?
+      !id.nil?
+    end
+
     protected
 
-    def add_accessor(symbol, *args)
+    def add_accessor(symbol, *_args)
```
backus: Can we just simplify this if we don't even use `_args`?

lib/blockscore/collection.rb
```diff
@@ -1,18 +1,239 @@
...
+    # @param [Class] class of collection members
+    #
+    # @return [undefined]
+    #
+    # @api private
+    def initialize(parent, member_class)
+      @parent       = parent
+      @member_class = member_class
+      register_parent_data
+    end
```
backus: I still think the collection class looks good, but I'm not sure if you saw my comment a day or two ago on hipchat after we worked on this.  Pasting it here:

> [Sep-8 6:53 PM] John Backus: @delmer what if we did something like this

```ruby
module BlockScore
  class Collection
    def initialize(parent)
      @parent = parent
      register_parent_data
    end

    def member_class
      self.class::MEMBER_CLASS
    end

    def default_params
      {
        self.class::PARENT_KEY => parent.id
      }
    end
  end
end

module BlockScore
  class Person
    class QuestionSets < Collection
      MEMBER_CLASS = QuestionSet
      PARENT_KEY = :person_id
    end
  end
end
```

lib/blockscore/collection.rb
```diff
@@ -1,18 +1,239 @@
...
+      end
+    end
+
+    # ids that belong to the collection
+    #
+    # @return [Array<String>]
+    #
+    # @api private
+    def ids
+      parent.attributes.fetch(:"#{Util.to_plural(member_class.resource)}", [])
```
backus: Looking at this again, I'm thinking `Util.to_plural(member_class.resource)` should just be its own method

lib/blockscore/collection/member.rb
```diff
@@ -0,0 +1,100 @@
...
+      # @return nil otherwise
+      #
+      # @api private
+      def save_parent
+        parent.save unless parent_saved?
+      end
+
+      # Check if parent is saved
+      #
+      # @return [Boolean]
```
backus: Woops. This isn't correct technically.  Pretty sure this is my fault

lib/blockscore/collection/member.rb
```diff
@@ -0,0 +1,100 @@
...
+      #
+      # @return [Array<String>]
+      #
+      # @api private
+      def ids
+        parent.attributes.fetch(:"#{Util.to_plural(instance.class.resource)}", [])
+      end
+    end
+  end
+end
```
backus: It looks like `Member` and `Collection` share several methods.  Might be worth extracting to a module that is mixed into both?

lib/blockscore/collection/member.rb
DelmerGA: What's wrong here?

backus: I just left this comment mainly to remind myself since I introduced this.  Just the comment.  Doesn't return a boolean it returns nil or a string

lib/blockscore/collection/member.rb
DelmerGA: I saw the code smell too, ruby tapas all over again. I felt like it might be abstracted into a relationship module.

backus:
In general we try to follow the best practices for rspec.  Mainly I'd recommend checking out https://github.com/howaboutwe/rspec-style-guide http://betterspecs.org/.  A few specific points I would point out:

1. Prefer short `describe`s, short `it`s, and long `context`s.  Usually `describe` and `it` will just be a constant name or a method description like `'.foo'` or `'#bar'`.  Same for `it` minus the constant.   This isn't a strict rule since sometimes it is hard to avoid, but in general this is preferable and it also factors into how mutant works with your specs
2. Similar to the note above, feel free to omit `it` descriptions when possible.  For example:

```ruby
describe '#all' do
  subject { collection.all  }

  it 'should return self' do
    is_expected.to be(collection)
  end
end
```

vs.

```ruby
describe '#all' do
  subject { collection.all }
  it { should be(collection) }
end
```

lib/blockscore/collection/member.rb
backus: Yeah I like that.

DelmerGA:
LGTM

PR #39, opened 2015-09-14
Fix person-questions_sets-array

DelmerGA:
Editing out comments


DelmerGA:
Files review for extraneous comments, LGTM.

PR #43, opened 2015-09-17
Feature/vcr based testing

### LINKED ISSUE #42: Specs could use VCR and less stubbing
Currently, request stubbing restricts testing certain relationships, but VCR would help by utilizing the actual API responses.

sirwolfgang:
- Updated supported ruby version to maintained versions
- Removed webmock based request stubbing and testing
- Created VCR based request replaying and testing [resolves #42]
- Updated Readme


spec/spec_helper.rb
```diff
@@ -1,116 +1,74 @@
...
   RSpec.configure do |config|
     config.include(FactoryGirl::Syntax::Methods)
-    config.include(WebMock::API)
 
-    config.before(:each) do
-      @api_stub = stub_request(:any, Spec::STUB_PATTERN)
-                  .with(headers: Spec::HEADERS)
-                  .to_return(&Spec.method(:webmock_handler))
+    config.before(:suite) do
+      BlockScore.api_key = ENV['BLOCKSCORE_TEST_KEY'] || 'BLOCKSCORE_TEST_KEY'
```
backus: Why does the API key fallback to 'BLOCKSCORE_TEST_KEY'?  We might as well throw an error.

backus: Also I'd write this

```ruby
ENV.fetch('BLOCKSCORE_TEST_KEY', 'BLOCKSCORE_TEST_KEY')
```

and if you agree with my previous comment you can then just drop the second arg

sirwolfgang: Its not an error. Its the matching filler text so that if you don't have a key, your request matches the URI sorted in the sanitized cassette. The tests run with or without a value key.

backus: (y). Consider pulling the two into constants even if they are the same value.  More intention revealing then

spec/spec_helper.rb
```diff
@@ -1,116 +1,74 @@
...
+    # Add VCR to all tests
+    config.around(:each) do |example|
+      options = example.metadata[:vcr] || { record: (ENV['VCR_RECORD_MODE'] ? ENV['VCR_RECORD_MODE'].to_sym : :none) }
+      if options[:record] == :skip
+        VCR.turned_off(&example)
+      else
+        name = example.metadata[:full_description].split(/\s+/, 2).join('/').underscore.gsub(/\./, '/').gsub(/[^\w\/]+/, '_').gsub(/\/$/, '')
+        VCR.use_cassette(name, options, &example)
+      end
+    end
```
backus: I added the separate `spec.rb` file in order to separate out some of our settings and behavior from the spec_helper.  I don't have a strong opinion about doing it that way, but I think it makes it more clear how specs are set up since you then end up using cleaned up methods and well named methods as opposed to inlined code that isn't as clear.  I think that would make sense here since there is a lot going on in these few lines and they affect every spec in the suite.

spec/support/vcr_helper.rb
```diff
@@ -0,0 +1,10 @@
+module VCRHelper
+  def self.spinlock_until(&_block)
+    try_count = 0
+    loop do
+      try_count += 1
+      break if try_count >= 30 || yield
+      sleep 1
+    end
+  end
+end
```
backus: Usually for something like this I reach for ruby's `redo` or `retry`.  In this case though I might write this like 

```ruby
module VCRHelper
  def self.spinlock_until(retries: 30)
    retries.times do
      return if yield
      sleep(1)
    end
  end
end
```

spec/unit/blockscore/candidate_spec.rb
```diff
@@ -1,46 +1,192 @@
...
+          address_city: Faker::Address.city,
+          address_country_code: Faker::Address.country_code
+        )
+      end
+
+      it { is_expected.to be_persisted }
+      it { expect(candidate.class).to be BlockScore::Candidate }
+    end
+
+    pending '.find' do
```
backus: Should this still be pending due to some API issue?

spec/unit/blockscore/candidate_spec.rb
```diff
@@ -1,46 +1,192 @@
...
+        BlockScore::Candidate.new(
+          ssn: '0000',
+          date_of_birth: Faker::Date.backward(365 * 100).to_s,
+          name_first: Faker::Name.first_name,
+          name_last: Faker::Name.last_name,
+          address_street1: Faker::Address.street_address,
+          address_city: Faker::Address.city,
+          address_country_code: Faker::Address.country_code
+        )
+      end
```
backus: Any reason you opt to not use the factories for this subject and the following?

spec/unit/blockscore/candidate_spec.rb
```diff
@@ -1,46 +1,192 @@
...
+
+      it { is_expected.to be_persisted }
```
backus: you can still use `should` in cases like this

spec/unit/blockscore/candidate_spec.rb
```diff
@@ -1,46 +1,192 @@
...
+      end
+    end
+
+    describe '.retrieve' do
+      let(:candidate_id)  { create(:candidate).id }
+      subject(:candidate) { BlockScore::Candidate.retrieve(candidate_id) }
+
+      it { is_expected.to be_persisted }
+      it { expect(candidate.name_first).not_to be_empty }
+      it { expect(candidate.class).to be BlockScore::Candidate }
```
backus: since we have `rspec-its` as a dependency you can always write this as 

```ruby
its(:class) { should be(BlockScore::Candidate) }
```

I'm a fan of this when it is possible

spec/unit/blockscore/candidate_spec.rb
```diff
@@ -1,46 +1,192 @@
...
-      include_context 'api request'
+      it { expect(candidate.name_last).to eq 'Smith' }
     end
 
-    describe '#delete' do
-      before { candidate.delete }
+    describe '#inspect' do
+      subject(:candidate_inspection) { create(:candidate).inspect }
+      it { expect(candidate_inspection.class).to be String }
+    end
```
backus: I usually tell my coverage tools to ignore `#inspect` methods I customize.  I'm fine with specifying a test, but in this case we might as well assert it matches some format instead of just asserting it is a string

spec/unit/blockscore/candidate_spec.rb
```diff
@@ -1,46 +1,192 @@
...
+      before do
+        candidate.search
 
-        include_context 'POST'
+        # ==================================================
+        # Micro Patch: Spinlock until delay job has processed.
+        #
+        # After a new recording, the empty responses should be edited out to
+        #   prevent baking in the sleeps, as they will be replayed in order.
+        VCRHelper.spinlock_until { candidate.hits.count != 0 }
```
backus: I like the use of the helper here.  This is much more out of the way than I expected.

spec/unit/blockscore/collection/member_spec.rb
```diff
@@ -0,0 +1,8 @@
+module BlockScore
+  class Collection
+    RSpec.describe Member do
+      describe '.save' do
+      end
+    end
```
backus: Any plans for this file?  Also, to save yourself from nesting hell, you can always write this like so:

```ruby
RSpec.describe Member, '.save' do
end
```

spec/unit/blockscore/collection_spec.rb
```diff
@@ -1,178 +1,48 @@
...
 
-    before do
-      Util::PLURAL_LOOKUP['fake_resource'] = 'fake_resources'
-      allow(member_class).to receive(:create) { create(:fake_member, parent_id: parent.id) }
-    end
-
-    describe '#all' do
-      subject { collection.all  }
+    it { expect(question_sets.empty?).to be true }
+    it { expect(question_sets.class).to be BlockScore::Collection }
```
backus: I'd also use `its` syntax for both of these

backus: I don't feel strongly either way here, but since the following line

```ruby
it { expect(question_sets.class).to be BlockScore::Collection }
```

shows up 5 times in this file my gut reaction would be to reach for `shared_context`

spec/unit/blockscore/company_spec.rb
```diff
@@ -1,4 +1,136 @@
...
+          tax_id: '000000000',
+          incorporation_country_code: Faker::Address.country_code,
+          incorporation_type: 'corporation',
+          address_street1: Faker::Address.street_address,
+          address_city: Faker::Address.city,
+          address_subdivision: Faker::Address.state_abbr,
+          address_postal_code: Faker::Address.postcode,
+          address_country_code: Faker::Address.country_code
+        )
+      end
```
backus: Same comment regarding `attributes_for`

spec/unit/blockscore/company_spec.rb
```diff
@@ -1,4 +1,136 @@
...
+            address_country_code: Faker::Address.country_code
+          )
+        end
+
+        it { is_expected.to be_persisted }
+        it { expect(company.class).to be BlockScore::Company }
+      end
+    end
+
+    pending '.find' do
```
backus: Same pending comment

spec/unit/blockscore/person_spec.rb
```diff
@@ -1,42 +1,180 @@
...
+        it { expect(person.class).to be BlockScore::Person }
+      end
+    end
+
+    describe '#invalid?' do
+      context 'valid person' do
+        subject(:person) { create(:valid_person) }
+
+        it { is_expected.to be_persisted }
+        it { expect(person.invalid?).to be false }
```
backus: might as well use the predicate inflection you used in the line above for `invalid?` as well

spec/unit/blockscore/question_set_spec.rb
```diff
@@ -1,42 +1,128 @@
 module BlockScore
   RSpec.describe QuestionSet do
-    let(:person) { create(:person_params) }
+    describe '.new' do
+      subject(:person) do
+        BlockScore::QuestionSet.new
+      end
```
backus: I assume this is left over from duplicating the person spec?  Might want to look through this and other files that you bootstrapped with other specs to make sure naming and what not is fine

spec/unit/blockscore/question_set_spec.rb
```diff
@@ -1,42 +1,128 @@
...
+          answers = []
+
+          question_set.questions.each do |question|
+            question.answers.each do |answer|
+              if ['309 Colver Rd', '812', 'Jasper', '49230', 'None Of The Above'].include? answer.answer
+                answers << { question_id: question.id, answer_id: (answer.id + 1) % 5 }
+                break
+              end
+            end
+          end
```
backus: Might want to reduce the duplication between the different contexts here.  Maybe just pull the common part into a module

backus:
Done with this pass @sirwolfgang. Pretty solid overall especially given how quickly you did all of this.  My main comment is just to reduce duplication between the specs.   Just tag me in a comment when you've updated things and want me to take another look.  I think we're good to merge after that

spec/spec_helper.rb
sirwolfgang: I moved everything related into a `vcr_setup.rb`

spec/support/vcr_helper.rb
sirwolfgang: :+1:

spec/unit/blockscore/candidate_spec.rb
sirwolfgang: `.find` is pending because its not implemented yet.

spec/unit/blockscore/collection/member_spec.rb
sirwolfgang: We currently do not use this feature, so I have no exposed way of testing it. This is here as a reminder to test it, when we extend from it.

spec/vcr_setup.rb
```diff
@@ -0,0 +1,41 @@
...
+          .gsub(%r(\/+), '/')
+          .strip
+          .tr(' ', '_')
+          .gsub(%r([/_ ]$), '')
+
+        VCR.use_cassette(name, options, &example)
+      end
+    end
+  end
+end
```
backus: The comment might have been dropped in the revisions, but the reason I recommended moving this stuff into its own module was so that the configuration blocks could be pulled out into methods and whatnot.

backus: I didn't go as far as I usually would with modifying this, but this is sort of what I had in mind:

```ruby
require 'vcr'
require 'rspec'

module BlockScore
  RECORD_MODE_KEY     = 'VCR_RECORD_MODE'
  ENVIRONMENT_API_KEY = 'BLOCKSCORE_TEST_KEY'
  DEFAULT_API_KEY     = 'BLOCKSCORE_TEST_KEY'

  extend self

  RSpec.configure do |config|
    config.before(:suite) do
      BlockScore.api_key = ENV.fetch(ENVIRONMENT_API_KEY, DEFAULT_API_KEY)
    end

    config.around(:each) do |example|
      record_option = ENV.fetch(RECORD_MODE_KEY, :none).to_sym
      options = example.metadata.fetch(:vcr, record: record_option)
      run_vcr_example(options, example)
    end
  end

  VCR.configure do |c|
    c.cassette_library_dir = 'spec/cassettes'
    c.hook_into :webmock
    c.configure_rspec_metadata!
    c.ignore_hosts 'codeclimate.com'

    c.filter_sensitive_data(DEFAULT_API_KEY) { ENV[ENVIRONMENT_API_KEY] }
  end

  def run_vcr_example(options, examlpe)
    if options.fetch(:record).equal?(:skip)
      VCR.turned_off(&example)
    else
      name = extract_name(example.metadata.fetch(:full_description))

      VCR.use_cassette(name, options, &example)
    end
  end

  private

  def extract_name(full_description)
    full_description
        .split(/\s+/, 2)
        .join('/')
        .underscore
        .gsub(/[.#?]/, '/')
        .gsub(%r(\/+), '/')
        .strip
        .tr(' ', '_')
        .gsub(%r([/_ ]$), '')
  end
end
```

backus:
Finished second pass @sirwolfgang.  Added one comment on something I think I didn't communicate well with the VCR configuration stuff.  Mainly a note for the future.  Pull request LGTM.

PR #49, opened 2016-01-10
Update 40

### LINKED ISSUE #40: BlockScore::Base should offer an update
`BlockScore::Base#update` could have example usage like follows:

``` ruby
person = Person.retrieve("some_id")
person.update(document_type: "ssn", document_value: "0001")
```

connorjacobsen:
Adds an `:update` method to `BlockScore::Base` to add more of an `ActiveRecord` feel to the library.

In addition to the `:update` method, a protected `:initialize_from` method was added. This method will reinitialize the attributes of a given object from a provided values hash. The Stripe version of this
method updates the dynamically defined accessors and getters based on the values have changed. I have not done that because I do not think the value of an attribute changing is a positive signal as to whether or not a getter/setter will be called for that attribute in the future.

Also adds VCR cassettes for the update action, and changes the relevant spec from pending to active.

Fixes #40.

Bump version to 4.3.0 (Omit this commit if desired)

Minor version change due to addition of backwards compatible `:update` method.

### DIFF
--- a/lib/blockscore/actions/update.rb
+++ b/lib/blockscore/actions/update.rb
@@ -13,6 +13,11 @@ module BlockScore
     #  foo.name_first = 'John'
     #  foo.save
     #  # => true
+    #
+    #  foo = Foo.new(name_first: 'John')
+    #  foo.update(name_first: 'Jane')
+    #  foo.name_first
+    #  # => 'Jane'
     module Update
       extend Forwardable
 
@@ -46,7 +51,7 @@ module BlockScore
 
       def update!(params)
         response = patch(update_url, params)
-        initialize_from(response.attributes)
+        self.attributes = response.attributes
       end
 
       # Filters out the non-updateable params.
--- a/lib/blockscore/actions/update.rb
+++ b/lib/blockscore/actions/update.rb
@@ -27,19 +27,38 @@ module BlockScore
 
       def_delegators 'self.class', :endpoint, :patch
 
+      # May be deprecated in future releases, prefer :update.
       def save!
         if persisted?
-          patch("#{endpoint}/#{id}", filter_params)
+          patch(update_url, filter_params)
           true
         else
           super
         end
       end
 
+      def update(params)
+        update!(params)
+        true
+      rescue
+        false
+      end
+
+      def update!(params)
+        response = patch(update_url, params)
+        initialize_from(response.attributes)
+      end
+
       # Filters out the non-updateable params.
       def filter_params
         # Cannot %i syntax, not introduced until Ruby 2.0.0
         attributes.reject { |key, _| PERSISTENT_ATTRIBUTES.include?(key) }
+      end
+
+      private
+      
+      def update_url
+        "#{endpoint}/#{id}"    
       end
     end
   end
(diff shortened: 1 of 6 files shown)


connorjacobsen:
I do not have permission to see what the Code Climate issue is, so I need some help there.

backus:
> I do not have permission to see what the Code Climate issue is, so I need some help there.

same as flagged by hound

lib/blockscore/actions/update.rb
```diff
@@ -27,20 +32,39 @@ module Update
...
           true
         else
           super
         end
       end
 
+      def update(params)
+        update!(params)
+        true
+      rescue
```
backus: We should not blindly rescue from exceptions

lib/blockscore/actions/update.rb
```diff
@@ -27,20 +32,39 @@ module Update
...
+        false
+      end
+
+      def update!(params)
+        response = patch(update_url, params)
+        self.attributes = response.attributes
+      end
```
backus: Any reason for an update bang method?

lib/blockscore/base.rb
```diff
@@ -90,6 +90,11 @@ def add_setter(symbol, *_args)
       end
     end
 
+    # Allow reinitialization of an object's attributes. In the future, if
+    # more work is done during reinitialization, may want to break this out
+    # into a :initialize_from method.
+    attr_writer :attributes
```
backus: Can probably nix this comment

lib/blockscore/actions/update.rb
```diff
@@ -27,20 +32,39 @@ module Update
 
       def_delegators 'self.class', :endpoint, :patch
 
+      # May be deprecated in future releases, prefer :update.
```
backus: Is there an issue that says this?

lib/blockscore/actions/update.rb
connorjacobsen: There is also a [save!](https://github.com/BlockScore/blockscore-ruby/blob/master/lib/blockscore/base.rb#L50) and `ActiveRecord` has a `save` and `save!`, so it was for consistency purposes, but I'm not attached to it.

lib/blockscore/actions/update.rb
connorjacobsen: Would you want it to just pass the exception on, or do you have something else in mind? 

My point of reference was: https://github.com/BlockScore/blockscore-ruby/blob/master/lib/blockscore/base.rb#L46

backus: Ah well that shouldn't have made it through review. I have [noted it before](https://github.com/BlockScore/blockscore-ruby/pull/20#discussion_r30776192) though

lib/blockscore/actions/update.rb
backus: Are you expecting that we throw an error? If not then yeah delete this method

lib/blockscore/actions/update.rb
backus: See comment on update. Do you know what errors you are expecting?

lib/blockscore/actions/update.rb
connorjacobsen: I think it is very possible that we could, sure. It could be any of them, but the most likely would be `APIError` or `InvalidRequestError`. I think it really just comes down to how you want the gem to behave: should the gem itself do the error handling in some way, or should it be passed onto the user and allow them to deal with the different errors however they see fit. I think you could make a strong argument either way.

Just let me know what approach you want to take.

PR #50, opened 2016-01-11
allow overriding the api key

herval:
Hello,

One requirement of a system I'm working on is being able to use test/prod at the same time (on the same Rails environment/server/process/what-have-you). GIven BlockScore's api key is currently a global, that's not possible using the client. 

My proposal here is pretty minimal - just allow overriding the api key on a given request. This is backwards compatible by default and wouldn't impact existing uses of the client.

Let me know your thoughts!

### DIFF
--- a/lib/blockscore/connection.rb
+++ b/lib/blockscore/connection.rb
@@ -27,10 +27,11 @@ module BlockScore
     end
 
     def request(method, path, params)
-      fail NoAPIKeyError, 'No API key was provided.' unless BlockScore.api_key
+      api_key = params.delete(:api_key) || BlockScore.api_key
+      fail NoAPIKeyError, 'No API key was provided.' unless api_key
 
       begin
-        response = execute_request(method, path, params)
+        response = execute_request(method, path, params, api_key)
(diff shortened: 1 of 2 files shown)


backus:
Thanks for the PR! I'll talk this change over with the team and do a review later today.

backus:
First of all, thanks again for the PR and sorry this design decision by the library has gotten in your way. I've finally found some time to review.

This change impacts the entire public interface of the library it seems. Now, if I do

```ruby
BlockScore::Person.create(api_key: 'blah blah')
```

The request uses `"blah blah"` as the api key. I understand this is by design for your use case, but consider if someone was not filtering parameters before using our gem. An end user of theirs could add the `api_key` parameter to a request and that would end up changing the api_key used to send off a request to api.blockscore.com.

While this by itself wouldn't be a security issue, it is certainly a change in behavior that would require a version bump and even then I think this would be too easy for people to overlook. For these reasons I don't think I can merge this type of change into master.

I hope your fork has fixed this use case for you. We do plan to address this issue soon and I will ping you when the client officially supports this usage.

PR #52, opened 2016-01-21
Add metrics support to CI build

backus:
- Updated devtools

- Updated rubocop configuration to reflect our preferences

  - Added excludes to rubocop config for current violations in our codebase

- Improved our Circle CI build matrix trick that takes advantage of us having many parallel containers

- Added rubocop only to CI build

- Added other devtools config files for use down the road

**Note**: I didn't add ruby 2.3.0 to the build because it looks like Circle does not support it yet.


backus:
@tjchambers and @dkubb could you both take a look at this PR before I merge it? I'd appreciate feedback on the circleci bash script if you have any notes, but I mainly would like both of you to check out the configuration so you can get an idea of our style preferences.

tjchambers:
I looked at it and read through it. I have nothing to offer or would criticize.

backus:
Thanks

dkubb:
@backus I'd recommend passing the shell scripts through shellcheck.net first. It'll point out lots of things; the "error code" it displays is a hyperlink to a more in-depth explanation for a specific flag.

circleci.sh
```diff
@@ -1,11 +1,32 @@
...
 
-if [[ "$1" = "build" ]]
-then
+if [ $CIRCLE_NODE_TOTAL -le $NUM_RUBIES ]; then
+  echo "Cannot test $NUM_RUBIES ruby versions and metrics on $CIRCLE_NODE_TOTAL containers"
+  exit 1
+fi
+
+# Exit early if there are more containers than ruby versions to test
+if [ $CIRCLE_NODE_INDEX -gt $(($NUM_RUBIES)) ]; then
```
dkubb: What's the reason for wrapping this in `$((...))`? You didn't do it on line 7, so I wondered if maybe it was left-over code.

backus: Yup leftover. Originally I was subtracting one.

config/rubocop.yml
```diff
@@ -1,23 +1,33 @@
...
-Style/TrailingBlankLines:
-  Enabled: false
 Style/ExtraSpacing:
   Enabled: false
 Style/StringLiterals:
   EnforcedStyle: single_quotes
 Metrics/LineLength:
   Max: 100
+  Exclude:
+    - 'spec/unit/blockscore/candidate_spec.rb'    # temporary
```
dkubb: This is just a personal preference, but I would typically sort lists like this alphabetically so that there's no question where new entries should be added if the list should grow.

backus: Good point. In this case I just wanted to get a passing build so it wouldn't be a pain for you and @tjchambers to work with when the code changes down the road. I can make that change though

dkubb:
@backus once you've passed the shell script through shellcheck.net, ping me via IM and I can do another pass if you like.

(pinging me here might not be the best since my github activity feed is too active for me to monitor; I typically have pages and pages each day in my history)

backus:
@dkubb made the changes shellcheck.net recommended

PR #54, opened 2016-02-04
Deny nil retrieve argument

tjchambers:
The intent of this PR is to not make someone who passes nil as a value have to chase down a method missing on NilClass error but rather to be specific about what was provided.

@dkubb ready for review.

### DIFF
--- a/lib/blockscore/actions/retrieve.rb
+++ b/lib/blockscore/actions/retrieve.rb
@@ -11,8 +11,12 @@ module BlockScore
     # => #<BlockScore::Person:0x007fe39c424410>
     module Retrieve
       module ClassMethods
+
+        RESOURCE_ID_FORMAT = /\A[a-f0-9]+\z/.freeze
+
         def retrieve(id, options = {})
           fail ArgumentError, 'ID must be supplied' if id.nil? || id.empty?
+          fail ArgumentError, 'ID is malformed' unless id =~ RESOURCE_ID_FORMAT
           req = ->() { get("#{endpoint}/#{id}", options) }
           new(id: id, &req)
         end
--- a/lib/blockscore/actions/retrieve.rb
+++ b/lib/blockscore/actions/retrieve.rb
@@ -12,7 +12,7 @@ module BlockScore
     module Retrieve
       module ClassMethods
         def retrieve(id, options = {})
-          fail ArgumentError if id.empty?
+          fail ArgumentError, 'ID must be supplied' if id.nil? || id.empty?
           req = ->() { get("#{endpoint}/#{id}", options) }
           new(id: id, &req)
         end
--- a/spec/cassettes/block_score/not_found_error/behaves_like_an_error.yml
+++ b/spec/cassettes/block_score/not_found_error/behaves_like_an_error.yml
@@ -2,7 +2,7 @@
 http_interactions:
 - request:
     method: get
-    uri: https://BLOCKSCORE_TEST_KEY:@api.blockscore.com/people/invalid_id
+    uri: https://BLOCKSCORE_TEST_KEY:@api.blockscore.com/people/abc123
     body:
       encoding: US-ASCII
       string: ''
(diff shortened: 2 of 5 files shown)


dkubb:
@tjchambers instead of pulling in a merge commit, I would suggest squashing these two commits using `git rebase -i master`. Make sure your `master` branch is up-to-date with the current upstream.

lib/blockscore/actions/retrieve.rb
```diff
@@ -12,7 +12,7 @@ module Actions
     module Retrieve
       module ClassMethods
         def retrieve(id, options = {})
-          fail ArgumentError if id.empty?
+          fail ArgumentError,'ID must be supplied' if id.nil? || id.empty?
```
dkubb: This is a tiny preference, but I would typically have a trailing space after the comma.

tjchambers: Ok - I agree. So what is the best procedure for me to do that now that I have an active PR:

- Switch to my branch
- Adjust the code and commit

Now do I need to squash the commits on my end?

dkubb: Instead of checking for what the value isn't (blacklisting), what about checking for what the value should be? (whitelisting)

You could do something like:

```ruby
fail ArgumentError, 'ID must be be a natural number' unless id.to_s =~ /\A\d+\z/
```

I would tend to factor the regexp out into a constant though. I might also look at the number specifically. Would `0000` be a valid id? Is `01`? I might start off with something strict and then loosen it over time rather than start off weaker and tighten it -- the former is much easier than the latter.

In this specific case I might even recommend something like: `/\A[1-9]\d*\z/`

This prevents numbers with leading zeros. I don't know if there's a maximum length or if it's unbounded, but that could be factored in, eg: `/\A[1-9]\d{,5}\z/`

Also note the usage of `\A` and `\z`. Lots of people in ruby-land will reach for `^` and `$` in their regexps. 99.9% of the time __this is a bug__, consider:

```
"a\n1\nb" =~ /^\d+$/
=> 2

"a\n1\nb" =~ /\A\d+\z/
=> nil
```

If you have validation code using `^` and `$` in the regexp, there's a good chance invalid data can sneak by. I have found this to be extremely common in ruby apps.

dkubb:
@tjchambers I have finished a pass over this code.

lib/blockscore/actions/retrieve.rb
tjchambers: I like the whitelisting approach and was savvy to the \A and \z already. However I need to check with @backus as whitelisting - while useful IMO - requires me to be privy to their id scheme. Example Id is "54d956c06230620003760500". The upside is that it is prechecked (perhaps also for length?) and rejected. Downside - at least one I can think of - is if the ID regex changes it also means an API code change.

tjchambers:
@dkubb I am unable so far to figure out a way with the current GIT branch stream to collapse the merge commit. I believe it has to do with how it looked before, but I would really appreciate you looking at my repo and giving me some insight into how to merge them. It is clear how there are 2 commits between `Blockscore/master` and this branches head. But that intermediate merge has 2 parents. ANy help would be much appreciated.

tjchambers:
Added regex test for properly formatted resource arguments.

@dkubb ready for your re-review

dkubb:
@tjchambers the code looks good to me. I'm ready to merge, but it looks like CircleCI is stuck. I've asked @backus for access to CI so I can look into what is causing the problem.

dkubb:
@tjchambers Instead of working from a fork, I'd recommend working with normal branches within the project. It'll make things easier for us to collaborate. Technically you could grant me access to your repo, and when I need to rebase changes on top of master, I could do so, but it's just an extra layer of indirection.

dkubb:
@tjchambers This code will need to be rebased on top of master before it can be merged. I would do it myself, but I don't have access to your repo. Let me know when you've been able to rebase it so that the "Merge pull request" button becomes visible.

On another PR, it showed CircleCI as being required yet it was "stuck", but I was still able to merge it. That's a bit odd to me, I'm not sure why that was the case -- other projects I'm in have `master` locked, and it's not possible to circumvent and merge unless all checks have completed successfully.

dkubb:
The most plausible explanation of why the build is stuck is that CircleCI doesn't know how to build the commits from Tim's fork. CodeClimate seems to have no trouble though.

This won't be an issue once we're working in the same repo using feature branches.

PR #55, opened 2016-02-06
Add access to internet when VCR is turned off

tjchambers:
If record -> :skip is used, then most tests failed because access to the backend server was disallowed by WebMock. This temporarily allows external access and then turns it off again after example runs.

This permits mutant to run against the live server, which is necessary to eliminate false alive mutations. This also allows one to periodically confirm the VCR tests suite, which in this case highlighted a test that is passing but should not be.

### DIFF
--- a/spec/vcr_setup.rb
+++ b/spec/vcr_setup.rb
@@ -31,7 +31,9 @@ module BlockScore
 
   def run_vcr_example(options, example)
     if options.fetch(:record).equal?(:skip)
+      WebMock.allow_net_connect!
       VCR.turned_off(&example)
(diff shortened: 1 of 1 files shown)


tjchambers:
@dkubb ready for review

dkubb:
I'm going to merge this for now because it moves us forward with mutant integration, but I'd like to take a bit of time to audit how vcr is used. Normally I set things up so that all I have to do is delete the "vcr cassettes" and new ones are generated to replace the old ones. I would periodically remove them to test the functionality is up-to-date, and would consider any breakage to be a sign of a bug either on the server or client side, or in the specs.

PR #56, opened 2016-02-06
Upgrade error testing

tjchambers:
The error tests were inconsistent and incomplete. I attempted to make the code and the tests more DRY and consistent, as well as really test the messages returned via API. 

Depending on what sequence the PR’s are merged I may have to tweak this PR when I rebase it. 

@dkubb ready for review

### DIFF
--- a/config/rubocop.yml
+++ b/config/rubocop.yml
@@ -19,15 +19,14 @@ Style/StringLiterals:
 Metrics/LineLength:
   Max: 100
   Exclude:
-    - 'lib/blockscore/connection.rb'              # temporary
-    - 'lib/blockscore/util.rb'                    # temporary
-    - 'spec/unit/blockscore/candidate_spec.rb'    # temporary
-    - 'spec/unit/blockscore/candidate_spec.rb'    # temporary
-    - 'spec/unit/blockscore/collection_spec.rb'   # temporary
-    - 'spec/unit/blockscore/company_spec.rb'      # temporary
-    - 'spec/unit/blockscore/person_spec.rb'       # temporary
-    - 'spec/unit/blockscore/question_set_spec.rb' # temporary
-    - 'spec/unit/blockscore/question_set_spec.rb' # temporary
+    - 'lib/blockscore/connection.rb'                              # temporary
+    - 'lib/blockscore/util.rb'                                    # temporary
+    - 'spec/unit/blockscore/candidate_spec.rb'                    # temporary
+    - 'spec/unit/blockscore/collection_spec.rb'                   # temporary
+    - 'spec/unit/blockscore/company_spec.rb'                      # temporary
+    - 'spec/unit/blockscore/person_spec.rb'                       # temporary
+    - 'spec/unit/blockscore/question_set_spec.rb'                 # temporary
+    - 'spec/unit/blockscore/errors/invalid_request_error_spec.rb' # temporary
 Style/DotPosition:
   EnforcedStyle: leading
 Style/CommentAnnotation:
--- a/lib/blockscore/errors/api_error.rb
+++ b/lib/blockscore/errors/api_error.rb
@@ -1,3 +1,5 @@
+require 'blockscore/errors/error'
+
 module BlockScore
   class APIError < Error
     attr_reader :http_status
@@ -15,19 +17,29 @@ module BlockScore
     # APIError - Indicates an error on the server side (HTTP 5xx)
     # AuthenticationError - Indicates an authentication error (HTTP 401)
     def initialize(response)
-      body = JSON.parse(response.body, symbolize_names: true)
+      @http_body = JSON.parse(response.body, symbolize_names: true)
 
-      @message = body[:error][:message]
+      @message = error_advice[:message]
       @http_status = response.code
-      @error_type = body[:error][:type]
-      @http_body = body
+      @error_type = error_advice[:type]
     end
 
     def to_s
-      status_string = @http_status ? "(Status: #{@http_status})" : ''
-      type_string = @error_type ? "(Type: #{@error_type})" : ''
-
       "#{type_string} #{@message} #{status_string}"
+    end
+
+    private
+
+    def error_advice
+      http_body.fetch(:error)
+    end
+
+    def type_string
+      "(Type: #{error_type})" if error_type
+    end
+
+    def status_string
+      "(Status: #{http_status})" if http_status
     end
   end
 end
--- a/lib/blockscore/errors/api_error.rb
+++ b/lib/blockscore/errors/api_error.rb
@@ -19,9 +19,9 @@ module BlockScore
     def initialize(response)
       @http_body = JSON.parse(response.body, symbolize_names: true)
 
-      @message = error_advice[:message]
+      @message = error_advice.fetch(:message)
       @http_status = response.code
-      @error_type = error_advice[:type]
+      @error_type = error_advice.fetch(:type)
     end
 
     def to_s
@@ -35,11 +35,11 @@ module BlockScore
     end
 
     def type_string
-      "(Type: #{error_type})" if error_type
+      "(Type: #{error_type})"
     end
 
     def status_string
-      "(Status: #{http_status})" if http_status
+      "(Status: #{http_status})"
     end
   end
 end
--- a/lib/blockscore/errors/invalid_request_error.rb
+++ b/lib/blockscore/errors/invalid_request_error.rb
@@ -15,15 +15,17 @@ module BlockScore
     # end
     def initialize(response)
       super
-      @param = @http_body[:error][:param]
+      @param = error_advice[:param]
     end
 
     def to_s
-      status_string = @http_status ? "(Status: #{@http_status})" : ''
-      type_string = @error_type ? "(Type: #{@error_type})" : ''
-      param_string = @param ? "(#{@param})" : ''
-
       "#{type_string} #{@message} #{param_string} #{status_string}"
+    end
+
+    private
+
+    def param_string
+      "(#{param})" if param
     end
   end
 end
--- a/spec/support/error_behavior.rb
+++ b/spec/support/error_behavior.rb
@@ -1,7 +1,7 @@
 RSpec.shared_context 'an error' do
   it do
     expect { subject.call }.to raise_error do |raised|
-      expect(raised).to be_a(described_class)
+      expect(raised).to be_an_instance_of(described_class)
       expect(raised.message).to eql(message)
     end
   end
--- a/spec/support/error_behavior.rb
+++ b/spec/support/error_behavior.rb
@@ -1,7 +1,8 @@
 RSpec.shared_context 'an error' do
   it do
-    should raise_error(described_class) do |raised|
-      expect(raised.message).to eq(message)
+    expect { subject.call }.to raise_error do |raised|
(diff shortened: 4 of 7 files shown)


tjchambers:
@dkubb what is the correct procedure when getting reek line length errors on spec comparisons? Should I add an exception for this inline in the code?

tjchambers:
@dkubb - I meant rubocop not reek. Also if you could help me understand what I should have done to prevent this sneaking through with devtools enabled I would appreciate it. I have never used devtools, and suspect that before during running specs, I should also run something else.

tjchambers:
@dkubb ready for review again

lib/blockscore/errors/api_error.rb
```diff
@@ -15,19 +17,29 @@ class APIError < Error
     # APIError - Indicates an error on the server side (HTTP 5xx)
     # AuthenticationError - Indicates an authentication error (HTTP 401)
     def initialize(response)
-      body = JSON.parse(response.body, symbolize_names: true)
+      @http_body = JSON.parse(response.body, symbolize_names: true)
 
-      @message = body[:error][:message]
+      @message = error_advice[:message]
```
dkubb: Is the `:message` key always supposed to be present at all times? If so then I would use `Hash#fetch` here.

However, mutant should've exposed this case automaticaly. Did you run mutant against this code?

lib/blockscore/errors/api_error.rb
```diff
@@ -15,19 +17,29 @@ class APIError < Error
...
       @http_status = response.code
-      @error_type = body[:error][:type]
-      @http_body = body
+      @error_type = error_advice[:type]
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/56/files#r52110298

lib/blockscore/errors/invalid_request_error.rb
```diff
@@ -15,15 +15,17 @@ class InvalidRequestError < APIError
     # end
     def initialize(response)
       super
-      @param = @http_body[:error][:param]
+      @param = error_advice[:param]
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/56/files#r52110298

spec/support/error_behavior.rb
```diff
@@ -1,7 +1,8 @@
 RSpec.shared_context 'an error' do
   it do
-    should raise_error(described_class) do |raised|
-      expect(raised.message).to eq(message)
+    expect { subject.call }.to raise_error do |raised|
+      expect(raised).to be_a(described_class)
```
dkubb: I would tend to use `be_instance_of` instead `be_a` or `be_kind_of`. It's a stronger check, and if it becomes untrue I would want to know.

lib/blockscore/errors/api_error.rb
tjchambers: Yes  - I ran mutant - and mutant suggests it, but the tests are untrustworthy. So I asked the question re is it always present in Hipchat and did not get an answer yet from @backus.

spec/support/error_behavior.rb
tjchambers: good advice.

lib/blockscore/errors/api_error.rb
```diff
@@ -15,19 +17,29 @@ class APIError < Error
...
     end
 
     def to_s
-      status_string = @http_status ? "(Status: #{@http_status})" : ''
-      type_string = @error_type ? "(Type: #{@error_type})" : ''
-
       "#{type_string} #{@message} #{status_string}"
```
dkubb: I realize this wasn't changed, but in the event `type_string` is nil, do you want the error message to have a leading space? I might be more inclined to write this as:

```ruby
def to_s
  [type_string, @message, status_string].compact.join(' ')
end
```

lib/blockscore/errors/invalid_request_error.rb
tjchambers: In this case I do NOT believe param is always present. That is just a strong inference.

lib/blockscore/errors/api_error.rb
tjchambers: Yes - ditto - I asked in Hipchat to confirm that.

lib/blockscore/errors/api_error.rb
```diff
@@ -15,19 +17,29 @@ class APIError < Error
...
     end
+
+    private
+
+    def error_advice
+      http_body.fetch(:error)
+    end
+
+    def type_string
+      "(Type: #{error_type})" if error_type
```
dkubb: I don't see any tests for the case where `error_type` is falsey.

lib/blockscore/errors/api_error.rb
```diff
@@ -15,19 +17,29 @@ class APIError < Error
...
+    end
+
+    def status_string
+      "(Status: #{http_status})" if http_status
```
dkubb: I don't see any tests for the case where `http_status` is falsey.

config/rubocop.yml
```diff
@@ -19,15 +19,14 @@ Style/StringLiterals:
 Metrics/LineLength:
   Max: 100
```
dkubb: Aside from the specs in the `Exclude` block, do you know what the max is? I might be inclined to tighten this a bit and see what the absolute maximum is; and then everytime you touch an area of the rubocop config attempt to raise the bar a tiny bit.

lib/blockscore/errors/api_error.rb
tjchambers: Yet again - I don't either so I asked in Hipchat - @backus?

lib/blockscore/errors/api_error.rb
tjchambers: Yet again - I don't either so I asked in Hipchat - @backus?

tjchambers: My expectation is I would drop the conditionals and that would resolve the string interpolation question.

lib/blockscore/errors/api_error.rb
backus: Hm I think you can safely assume `:message` is always present

config/rubocop.yml
tjchambers: Presently the current max is 125 characters in the longest line.

lib/blockscore/errors/api_error.rb
tjchambers: I will treat it as so and use fetch.

lib/blockscore/errors/api_error.rb
backus: Yeah I don't see why `http_status` wouldn't be present. For situations like this you probably don't need to check with me if the code seems obviously wrong.

lib/blockscore/errors/api_error.rb
```diff
@@ -15,19 +17,29 @@ class APIError < Error
     # APIError - Indicates an error on the server side (HTTP 5xx)
     # AuthenticationError - Indicates an authentication error (HTTP 401)
     def initialize(response)
-      body = JSON.parse(response.body, symbolize_names: true)
+      @http_body = JSON.parse(response.body, symbolize_names: true)
```
dkubb: I wonder if using `symbolize_names` is a good idea here. It can cause a memory leak since it dynamically creates `Symbol` objects which are never GC'd.

However, the problem with changing this is that the `http_body` attribute is exposed as part of the public API. If the keys suddenly changed to `String` objects it might break things for users. The main reason I bring this up now is for us to be aware of it in the future.

Technically it could be fixed by having `JSON.parse` return string keys, filter out all but the allowed keys given a known list. Then we (deep) symbolize the resulting Hash. This would be safe because it would keep the number of keys bounded to a known limit. I haven't looked deeply, maybe there's always only an `:errors` key, and it only has a `:message` and `:type` underneath it. I don't know if it's worth the trouble now though.

lib/blockscore/errors/api_error.rb
tjchambers: OK - I will use my inference, and then we can relax as needed.

lib/blockscore/errors/api_error.rb
backus: Same response as https://github.com/BlockScore/blockscore-ruby/pull/56#discussion_r52110552

lib/blockscore/errors/api_error.rb
tjchambers: Noted.

config/rubocop.yml
dkubb: Another strategy that works well with rubocop metrics is to set the limit to exactly what your ideal threshold would be (in my case I would pick 79 chars), and then add exceptions for anything that does not meet that. It means having more Exclude entries, but it means that all new code will immediately be flagged so you can deal with it right away. Basing thresholds off legacy code usually doesn't provide as nice of a feedback loop as you will often be fixing recent code when you begin to clean things up -- better to draw the line in the sand now and proceed with the higher standards than trying to fix things later.

lib/blockscore/errors/api_error.rb
dkubb: So if `http_status` is always present, then I wouldn't bother with the conditional. Also my comment about `#to_s` is probably invalid (assuming `#error_type` is always present).

tjchambers: Correct - I would drop the conditionals.

config/rubocop.yml
backus: I would be fine with you changing this to 80 as long as you add a note above the excludes that says don't bother updating those offending lines just for the sake of resolving line length violations. I think it is a good quality filter for new code and code that needs to be otherwise refactored but I don't want it to be viewed as a "TODO" like some other rubocop exclusions.

tjchambers: I like the strategy, but if we lower it below 100 we will have approx 20 more excludes, so I will leave this for a separate decision by @backus.

tjchambers: OK - I will do that as a separate PR.

backus: I assume your first response below mine came in before you read my message but just to make sure: yes additional excludes are fine

tjchambers: Yes - and I will add the additional excludes as a separate PR since they are not related to this PR specifically. I just added the excludes I needed for this PR.

dkubb: Agreed.

spec/unit/blockscore/errors/not_found_error_spec.rb
```diff
@@ -1,6 +1,9 @@
 module BlockScore
   RSpec.describe NotFoundError do
     subject { -> { BlockScore::Person.retrieve('invalid_id') } }
+    let(:message) do
+      '(Type: invalid_request_error) Person with id invalid_id could not be found (Status: 404)'
```
tjchambers: @backus one thing I noted here is that Not found errors still return a type of invalid_request_error. I was confused by that at first, but noted that it is aligned with the API docs, AND I confirmed it interactively against the live server. SO while the message says not found and status = 404 the type doesn't feel like it coincides. THis IS how it is presently implemented and that response IS from the backend.

dkubb:
@tjchambers also note that this branch will need to be rebased on top of master before it can be mergeable.

tjchambers:
@dkubb ready for re-review...

PR #57, opened 2016-02-07
Fix rubocop error

dkubb:
This branch fixes a small rubocop error that oddly-enough was not flagged in a feature branch, but was flagged after merging into master.


dkubb:
@tchambers Do you mind reviewing/merging this?

tjchambers:
Definitely should not have passed. My bad. Odd it wasn't caught.

Issue #46, opened 2015-09-17
Move syntax toward a more ActiveRecord style


sirwolfgang:
- Add `#update` [issue #40]
- Alias `#destroy` to `#delete`
- Create `.find` to wrap `.retrieve` and provide non lazy loaded error bubbling
- Update error handling [issue #45]

PR #58, opened 2016-02-07
Remove long-standing pending specs tied to issue #46 AR style impl.

tjchambers:
Remove (for now) the pending specs associated with implementing an ActiveRecord style set of class methods in issue #46. These will be re-installed and updated if and when that issue is implemented.


tjchambers:
@dkubb ready for review

PR #59, opened 2016-02-08
Optimize target line length

tjchambers:
Using rubocop this PR lowers the line length max target to 80 for new code. 

I took the approach of quickly looking at the violations and determining if they could be easily corrected - especially in code vs. specs, or if by correcting this would add to the maintainability or correctness. This was done as an option to just adding the exclusion to rubocop.

### DIFF
--- a/lib/blockscore/base.rb
+++ b/lib/blockscore/base.rb
@@ -4,6 +4,8 @@ module BlockScore
   class Base
     extend Connection
 
+    ABSTRACT_WARNING = 'Base is an abstract class, not an API resource'.freeze
+
     def initialize(options = {}, &block)
       @loaded = !(block)
       @proc = block
@@ -63,7 +65,7 @@ module BlockScore
     end
 
     def self.endpoint
-      fail NotImplementedError, 'Base is an abstract class, not an API resource' if equal?(Base)
+      fail NotImplementedError, ABSTRACT_WARNING if equal?(Base)
 
       "#{api_url}#{Util.to_plural(resource)}"
     end
--- a/lib/blockscore/collection/member.rb
+++ b/lib/blockscore/collection/member.rb
@@ -93,7 +93,16 @@ module BlockScore
       #
       # @api private
       def ids
-        parent.attributes.fetch(:"#{Util.to_plural(instance.class.resource)}", [])
+        parent.attributes.fetch(resource_name, [])
+      end
+
+      # Plural resource name as symbol
+      #
+      # @return Symbol
+      #
+      # @api private
+      def resource_name
+        Util.to_plural(instance.class.resource).to_sym
       end
     end
   end
--- a/lib/blockscore/connection.rb
+++ b/lib/blockscore/connection.rb
@@ -1,10 +1,11 @@
 module BlockScore
   module Connection
-    ACCEPT_HEADER =
-      "application/vnd.blockscore+json;version=#{BlockScore::VERSION.to_i}"
-    USER_AGENT =
-      "blockscore-ruby/#{BlockScore::VERSION} (https://github.com/BlockScore/blockscore-ruby)"
-    CONTENT_TYPE = 'application/json'
+    MAJOR_VERSION = BlockScore::VERSION.to_i
+    REPO          = 'https://github.com/BlockScore/blockscore-ruby'.freeze
+    API_VERSION   = "version=#{MAJOR_VERSION}".freeze
+    ACCEPT_HEADER = "application/vnd.blockscore+json;#{API_VERSION}".freeze
+    USER_AGENT    = "blockscore-ruby/#{BlockScore::VERSION} (#{REPO})".freeze
+    CONTENT_TYPE  = 'application/json'.freeze
 
     def get(path, params)
       request :get, path, params
@@ -26,8 +27,8 @@ module BlockScore
 
     def headers
       @headers ||= {
-        'Accept' => ACCEPT_HEADER,
-        'User-Agent' => USER_AGENT,
+        'Accept'       => ACCEPT_HEADER,
+        'User-Agent'   => USER_AGENT,
         'Content-Type' => CONTENT_TYPE
       }
     end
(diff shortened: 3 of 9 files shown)


tjchambers:
@dkubb ready for review.

lib/blockscore/base.rb
```diff
@@ -4,6 +4,8 @@ module BlockScore
   class Base
     extend Connection
 
+    ABSTRACT_WARNING = 'Base is an abstract class, not an API resource'
```
dkubb: I would recommend freezing constants, especially strings.

lib/blockscore/connection.rb
```diff
@@ -1,5 +1,11 @@
 module BlockScore
   module Connection
+    MAJOR_VERSION = BlockScore::VERSION.to_i
+    REPO          = 'https://github.com/BlockScore/blockscore-ruby'
+    ACCEPT_HEADER = "application/vnd.blockscore+json;version=#{MAJOR_VERSION}"
+    USER_AGENT    = "blockscore-ruby/#{BlockScore::VERSION} (#{REPO})"
+    CONTENT_TYPE  = 'application/json'
```
dkubb: The above 4 lines are instances of https://github.com/BlockScore/blockscore-ruby/pull/59/files#r52251279.

lib/blockscore/connection.rb
```diff
@@ -20,9 +26,9 @@ def delete(path, params)
 
     def headers
       @headers ||= {
-        'Accept' => "application/vnd.blockscore+json;version=#{BlockScore::VERSION.to_i}",
-        'User-Agent' => "blockscore-ruby/#{BlockScore::VERSION} (https://github.com/BlockScore/blockscore-ruby)",
-        'Content-Type' => 'application/json'
+        'Accept' => ACCEPT_HEADER,
+        'User-Agent' => USER_AGENT,
+        'Content-Type' => CONTENT_TYPE
```
dkubb: This is a small preference, but I would generally align the `=>` in a Hash assignment. I find it easier to read, and easier to edit.

The only downside is when you add a longer key, or remove the longest key and have to reformat the Hash, but that doesn't seem to be a big deal in practice.

lib/blockscore/util.rb
```diff
@@ -10,6 +10,10 @@ module Util
       'watchlist_hit' => 'watchlist_hits'
     }
 
+    PARSE_ERROR =
+      'An error has occurred. ' \
+        'If this problem persists, please message support@blockscore.com.'
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/59/files#r52251279

spec/support/question_set_helper.rb
```diff
@@ -1,5 +1,11 @@
 module QuestionSetHelper
-  VALID_ANSWERS = ['309 Colver Rd', '812', 'Jasper', '49230', 'None Of The Above']
+  VALID_ANSWERS = [
+    '309 Colver Rd',
+    '812',
+    'Jasper',
+    '49230',
+    'None Of The Above'
+  ]
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/59/files#r52251279

lib/blockscore/connection.rb
tjchambers: @backus I looked in your style guide. Normally I also freeze strings - especially on an API that won't soon be based on Ruby 3.x ONLY). But there was nothing in the style guide, pro or con, so I am going to follow @dkubb guidance, because that is what I also do.

spec/support/question_set_helper.rb
tjchambers: In my code I usually would apply IceNine, but it is not part of this API gem. SO I will freeze the strings and the array.

lib/blockscore/connection.rb
backus: Yup we prefer freezing, using IceNine and Adamantium, etc. In fact, I'd ask that you point out if dkubb's preferences contradict our style guide

tjchambers: At present there is no pref in your style guide, and this gemspec does not mention ice nine, adamantium, etc. DO you want them added where appropriate?

backus: Yup I don't think I want this gem to pull in dependencies I just wanted to point out we are fully aligned on preferences here.

tjchambers: Ok then will do so *manually*.

tjchambers:
I will work on this tomorrow and fix the Rubocop errors and add the freezing.

tjchambers:
@dkubb ready for another pass

lib/blockscore/collection/member.rb
```diff
@@ -93,7 +93,16 @@ def parent_saved?
       #
       # @api private
       def ids
-        parent.attributes.fetch(:"#{Util.to_plural(instance.class.resource)}", [])
+        parent.attributes.fetch(resource_name, [])
```
dkubb: Not something that needs to be resolve in this PR, but I wonder if we should freeze collections if modifying them doesn't result in changes being persisted via the API.

dkubb:
@tjchambers I'll merge this in now, but I noticed that each commit in this PR actually fails the build except for the last commit. I would typically suggest organizing your commits so that the code is always in a passing state; it helps with `git bisect` and just generally makes it easier to rearrange commits or cherry-pick them into other branches -- or (my favourite) splitting off commits from a large PR into smaller PRs that can be merged early while working on a larger change that is still in progress.

This is specifically what I noticed when sending all commits to the CI system individually:

screen shot 2016-02-09 at 7 55 50 pm

tjchambers:
I think I have 2 problems I am dealing with. The PR's are fine and each commit passes, until another PR is merged and then when rebased the current PR fails. I can't tell which PR's will get put in (in fact I didn't expect this one to be merged yet as I was not sure it would be ok AFTER the later numbered ones). Because so many are in the same code arena, there is lots of overlap. 

Second problem I have is after I rebase on master and push those, as you pointed out, the commits are not pushed individually. So I need to use the tool to push them out one at a time. 

The ultimate IMO is to have one commit per PR, but if I adhere to a simple SRP commit strategy that is a lot of PR's (dozens) and potentially lots more rebasing as well. 

I am not familiar with how to split off a PR once offered. Maybe there is a technique documented for that.

PR #60, opened 2016-02-08
Add test for Collection::Member class

tjchambers:
This originally had an erroneous describe block for .save rather than #save (instance method). In addition the test example block was empty (no test).


lib/blockscore/collection/member.rb
```diff
@@ -57,7 +57,7 @@ def parent_name
       #
       # @api private
       def save_parent
-        parent.save unless parent_saved?
+        parent.save! unless parent_saved?
```
dkubb: Should there be tests for both logic branches?

tjchambers:
This PR is on hold - do NOT merge.

spec/unit/blockscore/collection/member_spec.rb
```diff
@@ -1,7 +1,32 @@
...
+        end
+
+        context 'previously saved person' do
+          let(:person) { BlockScore::Person.create(attributes_for(:person)) }
+
+          it 'contains the proper fields' do
+            expect(member.save).to be true
+            expect(person).to be_persisted
+            expect(person.attributes.fetch(:question_sets)).to include question_set.id
+            expect(person.attributes.fetch(:question_sets).size).to be 1
```
dkubb: This isn't a blocker for this PR, but instead of these two matches I would probably just write:

```ruby
expect(person.attributes.fetch(:question_sets)).to eql([question_set.id])
```

If the question sets included anything different from what was expected the error message should include the diff of the data. No point in specifying two matches when one can do the same job.

spec/unit/blockscore/collection/member_spec.rb
```diff
@@ -1,7 +1,32 @@
...
+            expect(member).to be_an_instance_of described_class
+            expect(member.person_id).to eql person.id
```
dkubb: I think you could use `be(...)` here.

dkubb:
@tjchambers I realize this PR is on hold, but I wanted to let you know I attempted to rebase it on top of master (there was one small conflict) and the specs had one failure:

```
rspec ./spec/unit/blockscore/collection/member_spec.rb:21 # BlockScore::Collection::Member#save previously saved person contains the proper fields
```

spec/unit/blockscore/collection/member_spec.rb
```diff
@@ -1,7 +1,32 @@
 module BlockScore
   class Collection
```
mbj: If you put rspec descriptions into reopening the classes, you unintentionally make the potential private constants available to the examples. This can easily end up in testing against implementation details, not the public API.

For these, and the reason to remove indentation I suggest to kill this re-opening for all rspec tests.

I realize the whole code base had to be changed and this is out of scope for this PR. Commenting to start the discussion for a followup PR that fixes the entire class.

spec/unit/blockscore/collection/member_spec.rb
```diff
@@ -1,7 +1,32 @@
...
-      describe '.save' do
+      describe '#save' do
+        let(:question_set) { QuestionSet.new }
+        subject(:member) { described_class.new(person, question_set) }
+
+        context 'unsaved person' do
+          let(:person) { BlockScore::Person.new(attributes_for(:person)) }
+
+          it 'saves the person as part of member save' do
+            expect(person).to_not be_persisted
```
mbj: Does the `be_persisted` matcher test for booleans or truthyness, should there be a stronger matcher instead, or a more verbose expectation?

spec/unit/blockscore/collection/member_spec.rb
```diff
@@ -1,7 +1,32 @@
...
+            expect(member.save).to be true
+            expect(person).to be_persisted
+          end
+        end
+
+        context 'previously saved person' do
+          let(:person) { BlockScore::Person.create(attributes_for(:person)) }
+
+          it 'contains the proper fields' do
```
mbj: This example tests attributes individually. Resutling in the potential miss of specification when attributes are added in future.

What about to (assuming correct equalization exists) test against an expected instance. (+ the test of persistence as this is likely not part of the equalization behavior).

This would also eliminate the `be_an_instance_of` instance below.

Commenting just seeing the specs, assuming behavior of the class.

spec/unit/blockscore/collection/member_spec.rb
tjchambers: The be_persisted IMO should be .persisted?).to be true or false , and I have started the process by adjusting predicate methods to return true or false NOT nil. That process is not complete.

spec/unit/blockscore/collection/member_spec.rb
tjchambers: Thx. There is a much bigger problem in my opinion and that is that for `Person` there is a question_sets ivar AND there is an attribute[:question_sets] and they are maintained independently and NOT in sync. So access it one way and you get one thing (or nothing) and another and you get something that may be incomplete. It needs to be consistent, or access it the wrong way and you get confusion.

spec/unit/blockscore/collection/member_spec.rb
tjchambers: Thx. There is a lot lacking in this - hence why this is at best a very thin first attempt. It was better than `describe '.save' do ; end` but not by much. It was enough to discover loads of issues with `#save`.

spec/unit/blockscore/collection/member_spec.rb
tjchambers: @mbj Let me see if I understand:

Instead of layering the module, class references this should be:

`RSpec.describe BlockScore::Collection::Member ...`

spec/unit/blockscore/collection/member_spec.rb
tjchambers: @dkubb I thought `be` was equivalent to `equal`. That won't work with a string correct?

tjchambers:
@dkubb ready for another pass

spec/unit/blockscore/collection/member_spec.rb
dkubb: > Instead of layering the module, class references this should be:
>
> `RSpec.describe BlockScore::Collection::Member ...`

Yes, that is correct.

dkubb: @tjchambers given that this style was used in all the specs, I don't think this should be a blocker towards merging this pull request. However I would recommend a follow-up PR that fixes the specs to use the suggested approach, and then using it in all specs going forward.

dkubb:
@tjchambers I am going to merge this, but I would suggest doing a follow-up based on the recommendation in https://github.com/BlockScore/blockscore-ruby/pull/60#discussion_r52569347

PR #61, opened 2016-02-08
Correct candidate spec

tjchambers:
The force! method has no customer facing functionality and is being demoted to private.

The original test failed when VCR was skipped, due to the intent being that all retrievable objects are retrieved unless there is no ID match.


lib/blockscore/base.rb
```diff
@@ -92,6 +85,13 @@ def add_setter(symbol, *_args)
 
     private
 
+    def force!
+      res = @proc.call
+      @attributes = res.attributes.merge(@attributes)
+      @loaded = true
+      self
```
dkubb: Is this return value used anywhere? I can see where it could be used in the public interface, but the only caller I can see is calling the method in void context.

lib/blockscore/base.rb
```diff
@@ -92,6 +85,13 @@ def add_setter(symbol, *_args)
 
     private
 
+    def force!
+      res = @proc.call
```
dkubb: I wouldn't bother assigning to an lvar here and just use it inline on the line below.

spec/unit/blockscore/candidate_spec.rb
```diff
@@ -124,7 +124,7 @@ module BlockScore
       before { candidate.delete }
 
       it { is_expected.not_to be_persisted }
-      it { expect { BlockScore::Candidate.retrieve(candidate_id).force! }.to raise_error BlockScore::NotFoundError }
+      it { expect(candidate.deleted).to equal true }
```
dkubb: I would suggest using `be(true)` here.

tjchambers:
@dkubb ready for another pass when green

PR #62, opened 2016-02-08
Upgrade the consistency of the scoring interface

tjchambers:
This PR enforces the logic associated with the API Docs through specs which meet the following:

- If answers are provided to be scored they will be with a roundtrip to the server
- If malformed answers are provided they will also be passed to the server which should respond with InvalidRequestError
- If an empty set of answers are provided (proper format) then the score will be scored by server and return float 0.0
- If no answers are provided then the current score value is returned (nil if never scored, last score if previously scored question set)

Previously only correct answers and incorrect answers (both valid format) were tested completely.

### DIFF
--- a/lib/blockscore/question_set.rb
+++ b/lib/blockscore/question_set.rb
@@ -9,12 +9,14 @@ module BlockScore
     def_delegators 'self.class', :post, :endpoint
 
     def score(answers = nil)
-      if answers.nil? && attributes
-        attributes[:score]
-      else
-        @attributes = post("#{endpoint}/#{id}/score", answers: answers).attributes
-        score
-      end
+      rescore(answers) if answers
+      attributes.fetch(:score)
+    end
+
+    private
+
+    def rescore(answers)
+      @attributes = post("#{endpoint}/#{id}/score", answers: answers).attributes
     end
   end
 end
--- a/spec/unit/blockscore/question_set_spec.rb
+++ b/spec/unit/blockscore/question_set_spec.rb
@@ -24,7 +24,7 @@ module BlockScore
     end
 
     describe '.retrieve' do
-      let(:question_set_id)  { create(:question_set).id }
+      let(:question_set_id) { create(:question_set).id }
       subject(:question_set) { BlockScore::QuestionSet.retrieve(question_set_id) }
 
       it { is_expected.to be_persisted }
@@ -38,17 +38,30 @@ module BlockScore
       let(:incorrect_answers) { QuestionSetHelper.incorrect_answers(question_set.questions) }
 
       context 'correct answers' do
-        before { question_set.score(correct_answers) }
-
-        its(:score) { should eq 100.0 }
-        it { expect(question_set.score(correct_answers)).to eq 100.0 }
+        it { expect(question_set.score(correct_answers)).to be 100.0 }
       end
 
       context 'incorrect answers' do
-        before { question_set.score(incorrect_answers) }
+        it { expect(question_set.score(incorrect_answers)).to be 0.0 }
+      end
 
-        its(:score) { should eq 0.0 }
-        it { expect(question_set.score(incorrect_answers)).to eq 0.0 }
+      context 'malformed answers' do
+        it 'raises an error when answers are not an array of hashes' do
+          expect { question_set.score([]) }
+            .to raise_error InvalidRequestError, \
+                            '(Type: invalid_request_error) ' \
+                            'Answers should be an array of objects  (Status: 400)'
+        end
+      end
+
+      context 'previous answers' do
+        before { question_set.score(correct_answers) }
+
+        it { expect(question_set.score).to be 100.0 }
+      end
+
+      context 'previously no answers' do
+        it { expect(question_set.score).to be nil }
       end
     end
 
--- a/spec/cassettes/block_score/question_set/score/correct_answers.yml
+++ b/spec/cassettes/block_score/question_set/score/correct_answers.yml
@@ -255,4 +255,56 @@ http_interactions:
         Of The Above"}]}]}'
     http_version: 
   recorded_at: Fri, 18 Sep 2015 20:14:25 GMT
-recorded_with: VCR 2.9.3
+- request:
+    method: post
+    uri: https://BLOCKSCORE_TEST_KEY:@api.blockscore.com/question_sets/55fc70a0323833000300089c
+    body:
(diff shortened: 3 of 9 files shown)


tjchambers:
@dkubb @backus ready for review

tjchambers:
Note: This makes question set specs/code 100% mutant covered.

spec/unit/blockscore/question_set_spec.rb
```diff
@@ -54,17 +54,25 @@ module BlockScore
...
+      context 'previous answers' do
+        before { question_set.score(correct_answers) }
+
+        it { expect(question_set.score).to equal 100.0 }
+      end
 
-        its(:score) { should eq 0.0 }
-        it { expect(question_set.score(incorrect_answers)).to eq 0.0 }
+      context 'previously no answers' do
+        it { expect(question_set.score).to be_nil }
```
dkubb: I would tend to use `be(nil)` since it's a stronger match. `be_nil` only asserts the object returns `true` for `#nil?` which isn't quite as precise.

spec/unit/blockscore/question_set_spec.rb
```diff
@@ -54,17 +54,25 @@ module BlockScore
...
+      end
+
+      context 'malformed answers' do
+        it { expect { question_set.score([]) }.to raise_error InvalidRequestError }
+      end
+
+      context 'previous answers' do
+        before { question_set.score(correct_answers) }
+
+        it { expect(question_set.score).to equal 100.0 }
```
dkubb: As mentioned in another PR I would tend to use `be(100.0)` instead of `equal(100.0)`, if only for consistency with other matchers that I've used.

lib/blockscore/question_set.rb
```diff
@@ -9,12 +9,14 @@ class QuestionSet < Base
     def_delegators 'self.class', :post, :endpoint
 
     def score(answers = nil)
-      if answers.nil? && attributes
-        attributes[:score]
-      else
-        @attributes = post("#{endpoint}/#{id}/score", answers: answers).attributes
-        score
-      end
+      rescore(answers) if answers
```
dkubb: I really like this refactoring.

spec/unit/blockscore/question_set_spec.rb
```diff
@@ -54,17 +54,25 @@ module BlockScore
...
-        before { question_set.score(correct_answers) }
-
-        its(:score) { should eq 100.0 }
-        it { expect(question_set.score(correct_answers)).to eq 100.0 }
+        it { expect(question_set.score(correct_answers)).to equal 100.0 }
       end
 
       context 'incorrect answers' do
-        before { question_set.score(incorrect_answers) }
+        it { expect(question_set.score(incorrect_answers)).to equal 0.0 }
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/62/files#r52250764

spec/unit/blockscore/question_set_spec.rb
```diff
@@ -54,17 +54,25 @@ module BlockScore
       let(:incorrect_answers) { QuestionSetHelper.incorrect_answers(question_set.questions) }
 
       context 'correct answers' do
-        before { question_set.score(correct_answers) }
-
-        its(:score) { should eq 100.0 }
-        it { expect(question_set.score(correct_answers)).to eq 100.0 }
+        it { expect(question_set.score(correct_answers)).to equal 100.0 }
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/62/files#r52250764

spec/unit/blockscore/question_set_spec.rb
```diff
@@ -54,17 +54,25 @@ module BlockScore
...
       end
 
       context 'incorrect answers' do
-        before { question_set.score(incorrect_answers) }
+        it { expect(question_set.score(incorrect_answers)).to equal 0.0 }
+      end
+
+      context 'malformed answers' do
+        it { expect { question_set.score([]) }.to raise_error InvalidRequestError }
```
dkubb: Is there a specific error message you can match on, or do all `InvalidRequestError` instances use the same error message? If the error message can differ then I would typically assert it in the raise_error call.

tjchambers:
@dkubb ready for another pass when green

tjchambers:
@dkubb ready for another pass. 

I will declare I am reticent on one spec having to do named `previous answers`. 

I tried adding a procession of expects taking the question_set subject from correct answers -> refresh with no answers to prove prior score was captured -> rescore with incorrect answers -> refresh with no answers to prove score be 0.0. 

VCR seems to keep getting in the way because the requests for all of those need to be done inside ONE example, yet they are same requests for 2nd and 4th refresh. 

So while the test spec as written tests *something* it is IMO incomplete.

lib/blockscore/question_set.rb
backus: Yeah this is way better

spec/unit/blockscore/question_set_spec.rb
```diff
@@ -38,17 +38,30 @@ module BlockScore
       let(:incorrect_answers) { QuestionSetHelper.incorrect_answers(question_set.questions) }
 
       context 'correct answers' do
-        before { question_set.score(correct_answers) }
-
-        its(:score) { should eq 100.0 }
-        it { expect(question_set.score(correct_answers)).to eq 100.0 }
+        it { expect(question_set.score(correct_answers)).to be 100.0 }
```
mbj: I'd generally put the argument to rspec matchers in parentheses. Reading "2 whitespace in a row" always puts doubt in the mind of the reader, since the ruby syntax easily "bites" people in terms of association.

I realize the reminder of the code base might not follow my suggestion right now, so only changing the instances in this diff would increase entropy.

Just placing this note here to start a discussion potentially resulting in a global change, or rspec specific rubocop configuration to enforce a consistent style.

tjchambers: @mbj - I am not clear on what you mean by your comment, but it sounds valuable. Could you provide an example?

mbj: Instead of:

`it { expect(question_set.score(correct_answers)).to be 100.0 }`

I meant to use:

`it { expect(question_set.score(correct_answers)).to be(100.0) }`

tjchambers:
OK - that I understand - that is not the *common practice* in this code. But I do understand.

PR #63, opened 2016-02-09
Mix of code improvements

tjchambers:
This is a handful of code improvements re: using of ==, and removal/refactoring of complex code.

@dkubb ready for review

### DIFF
--- a/lib/blockscore/collection.rb
+++ b/lib/blockscore/collection.rb
@@ -134,8 +134,7 @@ module BlockScore
     # @api public
     def retrieve(id)
       each do |item|
-        next unless item.id == id
-        return item
+        return item if item.id.eql?(id)
       end
 
       instance = member_class.retrieve(id)
@@ -198,7 +197,7 @@ module BlockScore
     #
     # @api private
     def parent_id?(item)
-      parent.id && item.send(foreign_key) == parent.id
+      parent.id && item.send(foreign_key).eql?(parent.id)
     end
 
     # Register a resource in collection
--- a/lib/blockscore/connection.rb
+++ b/lib/blockscore/connection.rb
@@ -48,12 +48,12 @@ module BlockScore
     def execute_request(method, path, params)
       auth = { username: BlockScore.api_key, password: '' }
 
-      if method == :get
-        path = encode_path_params(path, params)
-        params = nil
-      else
-        params = params.to_json
-      end
+      params = if method.equal?(:get)
+                 path = encode_path_params(path, params)
+                 nil
+               else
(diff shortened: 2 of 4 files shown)


lib/blockscore/collection.rb
```diff
@@ -133,8 +133,7 @@ def create(params = {})
     # @api public
     def retrieve(id)
       each do |item|
-        next unless item.id == id
-        return item
+        return item if item.id.eql?(id)
```
dkubb: If `id` is an integer, then you could use: `item.id.equal?(id)` since it's a stronger match.

dkubb: I realize you're only modifying one part of this class, but I question the design of this collection class. Also, I know you didn't design this so I'm speaking to whomever designed it:

I don't think it should inherit from an Array; you now have an entire super-wide interface that your object needs to adhere to. If you ever want to refactor the object you have to re-implement the dozens of methods in Array that it implements. A much better idea for a collection class is to have a has-a relationship to an Array or Hash under the hood. You can still make it Enumerable as long as you design an `#each` method (I can help with this), and you get to control *exactly* what other methods you want to expose in your public interface.

Also, given that this method does a lookup based on an `id`, and it does so by looping over the entire collection points to the wrong data structure being used. What I would do is change this to have-a `Hash` underneath, where the key is the id, and the value is the object. You can lookup the item quickly using `Hash#fetch`, and still have `#each`:

```ruby
include Enumerable

# ...

def each(&block)
  return to_enum unless block
  items.each_value(&block)
  self
end
```

lib/blockscore/collection.rb
```diff
@@ -197,7 +196,7 @@ def new_member(instance, &blk)
     #
     # @api private
     def parent_id?(item)
-      parent.id && item.send(foreign_key) == parent.id
+      parent.id && item.send(foreign_key).eql?(parent.id)
```
dkubb: I would use `#equal?` here instead of `#eql?`

tjchambers: These id's are strings.

lib/blockscore/collection.rb
tjchambers: None of the id's in the system are integers, they are strings (similar to uuid's). So eql? is the best I know - correct?

tjchambers:
@dkubb rebased and ready for re-review

lib/blockscore/collection.rb
dkubb: Yes, `#eql?` is the best you can get in this case if the `id` are string objects.

PR #65, opened 2016-02-10
Confirm empty question sets collection established at Person init

tjchambers:
This spec change confirms person initialization also creates a question set collection array. This kills an alive mutant, and besides one other creation sequencing issue ensures Person class is mutation covered.

### DIFF
--- a/spec/unit/blockscore/person_spec.rb
+++ b/spec/unit/blockscore/person_spec.rb
@@ -1,10 +1,12 @@
 module BlockScore
   RSpec.describe Person do
     describe '.new' do
-      subject(:person) { BlockScore::Person.new(attributes_for(:person)) }
+      let(:attrs)      { attributes_for(:person) }
+      subject(:person) { BlockScore::Person.new(attrs) }
 
       it { is_expected.not_to be_persisted }
       its(:class) { should be BlockScore::Person }
+      its(:question_sets) { should be_an_instance_of BlockScore::Collection }
     end
 
     describe '.create' do


tjchambers:
@dkubb assuming green ready for your review

spec/unit/blockscore/person_spec.rb
```diff
@@ -1,10 +1,12 @@
 module BlockScore
   RSpec.describe Person do
     describe '.new' do
-      subject(:person) { BlockScore::Person.new(attributes_for(:person)) }
+      let(:attrs)      { attributes_for(:person) }
+      subject(:person) { BlockScore::Person.new(attrs) }
```
dkubb: The subject is evaluated lazily, so there's no particular need to define the `:attrs` let block before the subject.

Typically I will order my spec setup like:

* subject
* let
* before
* after
* it/its blocks

The subject is the object under test, so I think it's the most important thing in the test and what should appear first.

spec/unit/blockscore/person_spec.rb
```diff
@@ -1,10 +1,12 @@
...
 
       it { is_expected.not_to be_persisted }
       its(:class) { should be BlockScore::Person }
+      its(:question_sets) { should be_an_instance_of BlockScore::Collection }
```
dkubb: I might suggest a separate `describe '#question_sets'` block with this. I generally would have the `.new` describe block focus on describing the behaviour and side effects of the specified method. The fact that there's a `#question_sets` method available on the object isn't really pertinent to the constructor.

spec/unit/blockscore/person_spec.rb
```diff
@@ -1,10 +1,12 @@
 module BlockScore
   RSpec.describe Person do
```
dkubb: As described in another PR, I would probably not monkey patch `RSpec.describe` inside the BlockScore namespace. By including it inside this namespace you're inadvertently modifying the object under test.

Also if you use the suggested `RSpec.describe BlockScore::Person`, then you can replace the places within the spec that refer to `BlockScore::Person` directly with `describe_class`.

spec/unit/blockscore/person_spec.rb
tjchambers: Your sequence is contrary to the convention(?) in this code, and to my experience. I always put the lets first. @backus?

spec/unit/blockscore/person_spec.rb
tjchambers: @dkubb it is the constructor that creates the question_sets ivar. So that was my reasoning for having it here. Is that not considered?

spec/unit/blockscore/person_spec.rb
tjchambers: This was suggested also by @mbj, and I tend to use `described_class` myself, but that is not the convention of the specs to date. @backus?

spec/unit/blockscore/person_spec.rb
dkubb: I mainly provide this suggestion because it offers a consistent guide that can be followed. I have no problem if subject comes first as long as it's consistent throughout the code. I do personally find having `subject` first makes it much easier to find when looking for the object under test -- I view `let` blocks as kind of secondary; they are always subordinate to the subject.

spec/unit/blockscore/person_spec.rb
dkubb: > it is the constructor that creates the question_sets ivar

I think this is a valid reason for the test to exist. Please disregard my suggestion.

spec/unit/blockscore/person_spec.rb
dkubb: @backus I would probably recommend we change this globally across the specs. It both removes one source of possible indeterminism (RSpec modifying a class under test so it has a different interface/behaviour as in normal usage), and allows for code simplification through things like `describe_class`.

tjchambers: @backus I concur with @dkubb and would do this across the board as a separate PR. 

@dkubb is this mergeable now?

PR #66, opened 2016-02-10
Fix require sequence - version needed earlier

tjchambers:
VERSION is referenced earlier in the load sequence with prior changes. This commit moves the version module load earlier in the sequence to comply.


tjchambers:
@dkubb please review - this one is likely a priority

PR #67, opened 2016-02-10
Upgrade code improvements

tjchambers:
Refactor code to remove extra local variables, unnecessary tests, and strengthen the access to common values

@dkubb ready for review

### DIFF
--- a/lib/blockscore/actions/retrieve.rb
+++ b/lib/blockscore/actions/retrieve.rb
@@ -16,7 +16,7 @@ module BlockScore
         def retrieve(id, options = {})
           fail ArgumentError, 'ID must be supplied' if id.nil? || id.empty?
           fail ArgumentError, 'ID is malformed' unless id =~ RESOURCE_ID_FORMAT
-          req = ->() { get("#{endpoint}/#{id}", options) }
+          req = ->() { get(endpoint + id, options) }
           new(id: id, &req)
         end
       end
--- a/lib/blockscore/base.rb
+++ b/lib/blockscore/base.rb
@@ -32,7 +32,7 @@ module BlockScore
     end
 
     def refresh
-      capture_attributes(self.class.retrieve(id))
+      capture_attributes(retrieve(id))
 
       true
     rescue Error
--- a/lib/blockscore/base.rb
+++ b/lib/blockscore/base.rb
@@ -1,10 +1,12 @@
 require 'blockscore/connection'
+require 'uri'
 
 module BlockScore
   class Base
     extend Connection
 
     ABSTRACT_WARNING = 'Base is an abstract class, not an API resource'.freeze
+    API_URL = URI.parse('https://api.blockscore.com/').freeze
 
     def initialize(options = {}, &block)
       @loaded = !(block)
@@ -50,14 +52,10 @@ module BlockScore
       @resource ||= Util.to_underscore(to_s.split('::').last)
     end
 
-    def self.api_url
-      'https://api.blockscore.com/'
-    end
-
     def self.endpoint
       fail NotImplementedError, ABSTRACT_WARNING if equal?(Base)
 
-      "#{api_url}#{Util.to_plural(resource)}"
+      API_URL + Util.to_plural(resource)
     end
 
     def persisted?
--- a/lib/blockscore/base.rb
+++ b/lib/blockscore/base.rb
@@ -4,9 +4,11 @@ require 'uri'
 module BlockScore
   class Base
     extend Connection
+    extend Forwardable
 
     ABSTRACT_WARNING = 'Base is an abstract class, not an API resource'.freeze
-    API_URL = URI.parse('https://api.blockscore.com/').freeze
+
+    def_delegators 'self.class', :endpoint, :post, :retrieve
 
     def initialize(options = {}, &block)
       @loaded = !(block)
@@ -44,7 +46,7 @@ module BlockScore
     end
 
     def save!
-      capture_attributes(self.class.post(self.class.endpoint, attributes))
+      capture_attributes(post(endpoint, attributes))
       true
     end
 
@@ -55,7 +57,11 @@ module BlockScore
     def self.endpoint
       fail NotImplementedError, ABSTRACT_WARNING if equal?(Base)
 
-      API_URL + Util.to_plural(resource)
+      Pathname(Util.to_plural(resource))
+    end
+
+    def member_endpoint
+      endpoint + id
     end
 
     def persisted?
--- a/lib/blockscore/base.rb
+++ b/lib/blockscore/base.rb
@@ -15,11 +15,11 @@ module BlockScore
     def attributes
       return @attributes if @loaded
       force!
-      @attributes
+      attributes
     end
 
     def id
-      @attributes.fetch(:id, nil)
+      attributes.fetch(:id, nil)
     end
 
     def inspect
@@ -28,8 +28,7 @@ module BlockScore
     end
 
     def refresh
-      res = self.class.retrieve(id)
-      @attributes = res.attributes
+      capture_attributes(self.class.retrieve(id))
 
       true
     rescue Error
@@ -43,9 +42,7 @@ module BlockScore
     end
 
     def save!
-      response = self.class.post(self.class.endpoint, attributes)
-      @attributes = response.attributes
-
+      capture_attributes(self.class.post(self.class.endpoint, attributes))
       true
     end
 
@@ -64,10 +61,14 @@ module BlockScore
     end
 
     def persisted?
-      !id.nil? && !attributes[:deleted]
+      !id.nil? && !attributes.fetch(:deleted, false)
     end
 
     protected
+
+    def capture_attributes(source)
+      @attributes = source.attributes
+    end
 
     def add_accessor(symbol, *_args)
       singleton_class.instance_eval do
--- a/lib/blockscore/base.rb
+++ b/lib/blockscore/base.rb
@@ -23,7 +23,7 @@ module BlockScore
     end
 
     def id
-      attributes.fetch(:id, nil)
+      attributes[:id]
     end
 
     def inspect
@@ -65,10 +65,10 @@ module BlockScore
     end
 
     def persisted?
-      !id.nil? && !attributes.fetch(:deleted, false)
+      !id.nil? && !deleted?
     end
 
-    protected
+    private
 
     def capture_attributes(source)
       @attributes = source.attributes
@@ -90,7 +90,9 @@ module BlockScore
       end
     end
 
-    private
+    def deleted?
+      attributes.fetch(:deleted, false)
+    end
 
     def force!
       @attributes = @proc.call.attributes.merge(@attributes)
--- a/lib/blockscore/candidate.rb
+++ b/lib/blockscore/candidate.rb
@@ -8,7 +8,7 @@ module BlockScore
     include BlockScore::Actions::Delete
     include BlockScore::Actions::All
 
-    def_delegators 'self.class', :endpoint, :get, :post
+    def_delegators 'self.class', :get, :post
 
     def history
       resource_member 'history'
@@ -19,17 +19,13 @@ module BlockScore
     end
 
     def search(options = {})
-      post uri, options.merge(candidate_id: id)
-    end
-
-    def uri
-      API_URL + 'watchlists'
+      post 'watchlists', options.merge(candidate_id: id)
     end
 
     private
 
-    def resource_member(member)
-      get "#{endpoint}/#{id}/#{member}", {}
+    def resource_member(target)
+      get member_endpoint + target, {}
     end
   end
 end
--- a/lib/blockscore/candidate.rb
+++ b/lib/blockscore/candidate.rb
@@ -8,7 +8,7 @@ module BlockScore
     include BlockScore::Actions::Delete
     include BlockScore::Actions::All
 
-    def_delegators 'self.class', :api_url, :endpoint, :get, :post
+    def_delegators 'self.class', :endpoint, :get, :post
 
     def history
       resource_member 'history'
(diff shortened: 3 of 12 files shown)


lib/blockscore/base.rb
```diff
@@ -5,6 +5,7 @@ class Base
     extend Connection
 
     ABSTRACT_WARNING = 'Base is an abstract class, not an API resource'.freeze
+    API_URL = 'https://api.blockscore.com/'.freeze
```
mbj: Maybe you should make this an instance of `Addressable::URI`, this is like `Pathname` for file paths a much better abstraction for the concept of URI than a string.

lib/blockscore/base.rb
```diff
@@ -53,22 +51,22 @@ def self.resource
...
 
-    def self.api_url
-      'https://api.blockscore.com/'
-    end
-
     def self.endpoint
       fail NotImplementedError, ABSTRACT_WARNING if equal?(Base)
 
-      "#{api_url}#{Util.to_plural(resource)}"
+      "#{API_URL}#{Util.to_plural(resource)}"
```
mbj: If you follow the idea from https://github.com/BlockScore/blockscore-ruby/pull/67/files#diff-d64d05c8aeab202e4a33798a74a3eed0R8 you might be able to replace this with `API_URL.merge(path: Util.to_plural(resource)` avoiding string interpolation.

lib/blockscore/candidate.rb
```diff
@@ -19,7 +19,7 @@ def hits
     end
 
     def search(options = {})
-      post "#{api_url}watchlists", options.merge(candidate_id: id)
+      post "#{API_URL}watchlists", options.merge(candidate_id: id)
```
mbj: Instance of https://github.com/BlockScore/blockscore-ruby/pull/67/files#r52547249.

lib/blockscore/base.rb
tjchambers: @mbj thanks for your insights, I should have thought of that myself as I did the same change on other code a week ago. Changed.

lib/blockscore/question_set.rb
```diff
@@ -16,7 +16,7 @@ def score(answers = nil)
     private
 
     def rescore(answers)
-      @attributes = post("#{endpoint}/#{id}/score", answers: answers).attributes
+      capture_attributes(post("#{endpoint}/#{id}/score", answers: answers))
```
mbj: I wounder if it might be possible to use a `Pathname` instance to build the path from.

`endpoint.join(id, 'score')` would read much nicer than the string interpolation. Also Pathname provides a much more narrow interface than a `String` instance.

lib/blockscore/base.rb
```diff
@@ -1,10 +1,12 @@
 require 'blockscore/connection'
+require 'uri'
 
 module BlockScore
   class Base
     extend Connection
 
     ABSTRACT_WARNING = 'Base is an abstract class, not an API resource'.freeze
+    API_URL = URI.parse('https://api.blockscore.com/').freeze
```
dkubb: How do you feel about using `Addressable::URI` instead of the built-in `URI` class? The API is cleaner, it's much better tested, and it handles cases where `URI` falls over.

lib/blockscore/base.rb
```diff
@@ -15,11 +17,11 @@ def initialize(options = {}, &block)
     def attributes
       return @attributes if @loaded
       force!
-      @attributes
+      attributes
     end
 
     def id
-      @attributes.fetch(:id, nil)
+      attributes.fetch(:id, nil)
```
dkubb: Given that you just return `nil` in the event the id is missing, I would use `Hash#[]` here instead.

mbj: This could actually be a mutation:

```
foo.fetch(bar, nil) -> foo[bar]
```

lib/blockscore/base.rb
```diff
@@ -53,22 +52,22 @@ def self.resource
...
     def self.endpoint
       fail NotImplementedError, ABSTRACT_WARNING if equal?(Base)
 
-      "#{api_url}#{Util.to_plural(resource)}"
+      API_URL + Util.to_plural(resource)
     end
 
     def persisted?
-      !id.nil? && !attributes[:deleted]
+      !id.nil? && !attributes.fetch(:deleted, false)
```
dkubb: I'd recommend moving `attributes.fetch(:deleted, false)` inside a private method named `#deleted?` and then change this to:

```ruby
def persisted?
  !id.nil? && !deleted?
end
```

lib/blockscore/base.rb
mbj: Or potentially more generalized:

```
foo.fetch(bar, baz) -> foo[bar]
```

This will catch cases where the explicitly provided default equals the implicit one.

lib/blockscore/base.rb
```diff
@@ -53,22 +52,22 @@ def self.resource
...
     end
 
     protected
```
dkubb: What is `protected` being used for in this case? In ruby `protected` is usually not what people think it is.

All it does is provides an API that other objects of the same type can use to access internal state. It's useful in things like `#eql?` or `#==` where you want to compare against some other object but the public API doesn't give you access to the internal state you need.

Most of the time when people use `protected` in ruby they have a different understanding of what it's doing, and it is probably a bug except when it satisfies a very narrow use case.

lib/blockscore/base.rb
mbj: Should this than better be written as:

```ruby
def persisted?
  !(id.nil? || deleted?)
end
```

Since it is equivalent under demorgans law and has less AST nodes, at equivalent maximal depth.

lib/blockscore/base.rb
dkubb: > This could actually be a mutation:
> 
> ```ruby
> foo.fetch(bar, nil) -> foo[bar]
> ```

Yes, exactly. This is a common scenario. While I prefer `Hash#fetch` for most things, I also believe that if your use case *exactly* corresponds to the API of a given method, you should use that method rather than recreating it's semantics using something else.

lib/blockscore/base.rb
dkubb: @mbj I thought about that, but I find `!(id.nil? || deleted?)` to be more difficult to understand.

The original example I read as "id is not nil and it's not deleted", while I felt like I was performing a few extra brain cycles to understand the second version.

lib/blockscore/base.rb
mbj: Yes. If you want a default to `nil` (or something not dynamically computed already present in the instances `Hash#default`) you want `Hash#[]`.

lib/blockscore/base.rb
mbj: I think its reasonable to derivate from least complex AST here.

dkubb: Yeah, AST complexity is a heuristic we use *when all other things are equal* to determine a choice between multiple equivalent pieces of code. However if one is easier to understand than the other, then it's not equal, it's superior to the other choices, and should be ranked higher.

lib/blockscore/question_set.rb
tjchambers: I hadn't thought of using `Pathname` because it wasn't a `file` path, but with endpoint being a URI that seems like a suitable alternative.

lib/blockscore/base.rb
tjchambers: I was cautioned not to include other gems, so I chose to use what was readily available and backwards compatible with the Ruby version supported. The more dependencies on other gems the more possibility also of issues with Ruby version support conflicts. After I looked at `Addressable::URI` I saw no distinct advantage.

lib/blockscore/base.rb
tjchambers: I will change it to use Hash#[] form.

lib/blockscore/base.rb
tjchambers: I will change.

lib/blockscore/base.rb
tjchambers: I will see how `protected`plays into this code. It is rarely used. One of my concerns which I have not addressed is the overall landscape of methods that are publicly accessible. It feels like it is too broad.

lib/blockscore/base.rb
dkubb: > The more dependencies on other gems the more possibility also of issues with Ruby version support conflicts

This is a valid concern.

> After I looked at Addressable::URI I saw no distinct advantage.

It handles more valid kinds of URIs than the normal URI class. The URI class has been basically frozen in time for the last 5-10 years and imho has been poorly maintained, as has much of ruby's core/stdlib. There are in-the-wild URIs that it cannot handle. The addressable gem more closely adheres to modern RFCs.

The API is mostly compatible, so refactoring to it later would not be much of an issue. If we run into actual issues that a modern URI replacement could resolve then we can revisit this.

tjchambers: In a soon-to-be-committed change you will see I have isolated this URI code to one module. So it should be very easy to swap out if needed.

tjchambers:
@dkubb when green ready for another pass

lib/blockscore/actions/retrieve.rb
```diff
@@ -16,7 +16,7 @@ module ClassMethods
         def retrieve(id, options = {})
           fail ArgumentError, 'ID must be supplied' if id.nil? || id.empty?
           fail ArgumentError, 'ID is malformed' unless id =~ RESOURCE_ID_FORMAT
-          req = ->() { get("#{endpoint}/#{id}", options) }
+          req = ->() { get(endpoint + id, options) }
```
dkubb: Can this use `member_endpoint` too?

lib/blockscore/base.rb
```diff
@@ -53,21 +54,25 @@ def self.resource
...
-
     def self.endpoint
       fail NotImplementedError, ABSTRACT_WARNING if equal?(Base)
 
-      "#{api_url}#{Util.to_plural(resource)}"
+      Pathname(Util.to_plural(resource))
+    end
+
+    def member_endpoint
+      endpoint + id
```
dkubb: I just wanted to confirm, because I didn't see any specs, but does this properly add a slash in between the `endpoint` and `id`? I looked at the implementation of the `endpoint` class method and I didn't get any clues whether this could be the case.

lib/blockscore/candidate.rb
```diff
@@ -19,13 +19,13 @@ def hits
...
-      post "#{api_url}watchlists", options.merge(candidate_id: id)
+      post 'watchlists', options.merge(candidate_id: id)
     end
 
     private
 
-    def resource_member(member)
-      get "#{endpoint}/#{id}/#{member}", {}
+    def resource_member(target)
+      get member_endpoint + target, {}
```
dkubb: I also want to comfirm that appending the `target` to `member_endpoint` is going to result in the expected URL?

lib/blockscore/actions/retrieve.rb
tjchambers: @dkubb no because this is a class method not an instance method.

lib/blockscore/base.rb
tjchambers: Yes - the tests all pass, and in my RSpec refactoring there will be specific tests in Base. The endpoint is a Pathname so adding segments joins with the slash.

lib/blockscore/candidate.rb
tjchambers: Yes - the tests all pass, and in my RSpec refactoring there will be specific tests in `Base`. The endpoint is a `Pathname` so adding segments returns another `Pathname` which joins with the slash.

tjchambers:
@dkubb ready for another pass

lib/blockscore/actions/retrieve.rb
```diff
@@ -16,7 +16,7 @@ module ClassMethods
...
           new(id: id, &req)
```
dkubb: In a future refactoring I would suggest moving the line above into this line so it becomes:

```ruby
new(id: id) { get(endpoint + id, options) }
```

There doesn't seem to be any point in defining a separate lambda.

Also, you could yield the object into the block without breaking the public API (because anyone using this api now would not be yielding anything to the block, which would not raise an error). This would allow you to use `#member_endpoint`, since the `id` would've been set in the constructor. So this could then become:

```ruby
new(id: id) { |object| get(object.member_endpoint, options) }
```

tjchambers: That is a construction I was not aware ot. This is a good learning experience for me.

dkubb: > That is a construction I was not aware ot

Yeah, it's not *exactly* as straight-forward as this, but it's pretty close:

```diff
diff --git a/lib/blockscore/actions/retrieve.rb b/lib/blockscore/actions/retrieve.rb
index 5aac386..9d1c7e6 100644
--- a/lib/blockscore/actions/retrieve.rb
+++ b/lib/blockscore/actions/retrieve.rb
@@ -16,8 +16,7 @@ module BlockScore
         def retrieve(id, options = {})
           fail ArgumentError, 'ID must be supplied' if id.nil? || id.empty?
           fail ArgumentError, 'ID is malformed' unless id =~ RESOURCE_ID_FORMAT
-          req = ->() { get(endpoint + id, options) }
-          new(id: id, &req)
+          new(id: id) { |object| get(object.member_endpoint, options) }
         end
       end
 
diff --git a/lib/blockscore/base.rb b/lib/blockscore/base.rb
index e2cddad..13c14e2 100644
--- a/lib/blockscore/base.rb
+++ b/lib/blockscore/base.rb
@@ -95,8 +95,9 @@ module BlockScore
     end
 
     def force!
-      @attributes = @proc.call.attributes.merge(@attributes)
       @loaded = true
+      args = @proc.arity.zero? ? [] : [self]
+      @attributes = @proc.call(*args).attributes.merge(@attributes)
     end
 
     def method_missing(method, *args, &block)
```

The reason for moving `@loaded` earlier is to break an infinite loop that can occur when the block references `attributes` (which are not yet loaded, which triggers `#force!`, etc)

PR #68, opened 2016-02-12
Refactor RSpec arrangement and adjust namespaces

tjchambers:
- Refactor the headers in the unit tests to combine the module BlockScore into the Spec describe block to not pollute the underlying namespace
- The above change has two effects 1) dedenting the code (making it all marked as changed), and 2) requiring a correct directory name block_score not blockscore
- Thus all the unit test moved to unit/block_score and now all the files are deleted and added
- This had the side effect of meaning that the cassettes were also regenerated in most cases under a new directory path
- The errors which were arbitrarily under an errors sub-directory are now under block_score directory
- the above necessitated changing the exclusions in rubocop.yml but otherwise satisfied naming conventions in rubocop.

In addition to these global changes, there are many internal changes:

- spec code was shortened to eliminate max line length faults (BTW that means that rubocop exclusions need to be revisited - they are not automatically dropped)
- all cases of ' eq ' tests where promoted to eql or equal to strengthen assertions
- all be_a comparisons were strengthened to an appropriate be_an_instance_of
- described_class was inserted everywhere appropriate

Also some tests from the various member classes where pushed into a new Base_spec.rb as that is where the unit tests should reside (nb: base_spec.rb is incomplete and will be addressed in short-term future)

What is not done?

The cases of be_persisted, etc. should be changed to persisted? be true to strengthen the assertion of predicate methods returning true or false.
And of course once this is in place than another pass to ensure method coverage.
Addition of specific unit tests for the actions is also a to-do.

### DIFF
--- a/spec/unit/block_score/api_connection_error_spec.rb
+++ b/spec/unit/block_score/api_connection_error_spec.rb
@@ -2,7 +2,7 @@ RSpec.describe BlockScore::APIConnectionError do
   let(:stubbed_route_pattern) { %r{.*api\.blockscore\.com/people/abc123} }
   let(:stubbed_error) { Errno::ECONNREFUSED }
   let(:message) { 'Connection refused - Exception from WebMock' }
-  subject { -> { BlockScore::Person.retrieve('abc123') } }
+  subject { BlockScore::Person.retrieve('abc123') }
 
   before { stub_request(:get, stubbed_route_pattern).to_raise(stubbed_error) }
 
--- a/spec/unit/block_score/api_connection_error_spec.rb
+++ b/spec/unit/block_score/api_connection_error_spec.rb
@@ -1,12 +1,10 @@
-module BlockScore
-  RSpec.describe APIConnectionError do
-    let(:stubbed_route_pattern) { %r{.*api\.blockscore\.com/people/abc123} }
-    let(:stubbed_error) { Errno::ECONNREFUSED }
-    let(:message) { 'Connection refused - Exception from WebMock' }
-    subject { -> { BlockScore::Person.retrieve('abc123') } }
+RSpec.describe BlockScore::APIConnectionError do
+  let(:stubbed_route_pattern) { %r{.*api\.blockscore\.com/people/abc123} }
+  let(:stubbed_error) { Errno::ECONNREFUSED }
+  let(:message) { 'Connection refused - Exception from WebMock' }
+  subject { -> { BlockScore::Person.retrieve('abc123') } }
 
-    before { stub_request(:get, stubbed_route_pattern).to_raise(stubbed_error) }
+  before { stub_request(:get, stubbed_route_pattern).to_raise(stubbed_error) }
 
-    it_behaves_like 'an error'
-  end
+  it_behaves_like 'an error'
 end
--- a/spec/unit/block_score/candidate_spec.rb
+++ b/spec/unit/block_score/candidate_spec.rb
@@ -1,137 +1,121 @@
 require 'faker'
 
-module BlockScore
-  RSpec.describe Candidate do
-    describe '.new' do
-      subject(:candidate) { BlockScore::Candidate.new(attributes_for(:candidate)) }
+RSpec.describe BlockScore::Candidate do
+  describe '.new' do
+    subject(:candidate) { described_class.new(attributes_for(:candidate)) }
 
-      it { is_expected.not_to be_persisted }
-      its(:class) { should be BlockScore::Candidate }
+    it { is_expected.not_to be_persisted }
+    it { is_expected.to be_an_instance_of described_class }
+  end
+
+  describe '.create' do
+    subject(:candidate) { described_class.create(attributes_for(:candidate)) }
+
+    it { is_expected.to be_persisted }
+    it { is_expected.to be_an_instance_of described_class }
+  end
+
+  describe '.retrieve' do
+    let(:candidate_id) { create(:candidate).id }
+    subject(:candidate) { described_class.retrieve(candidate_id) }
+
+    it { is_expected.to be_persisted }
+    its(:name_first) { is_expected.not_to be_empty }
+    it { is_expected.to be_an_instance_of described_class }
+  end
+
+  # NB: At the time of writing, the testing API does not reset. Making this
+  #     test less then perfect. To work around the near certain availability
+  #     of historical candidates from other tests known candidates are created
+  #     and checked. This has the side effect of being susceptible to race
+  #     conditions if tested against the live API in parallel.
+  describe '.all' do
+    let(:uniq_token_one) { '05816347d2234ef4a92465e9784c7ce1' }
+    let(:uniq_token_two) { 'b4cd5fb645788c3b37c11407b83d7741' }
+    subject(:candidates) { described_class.all }
+
+    before do
+      create(:candidate, name_first: 'John', name_last: uniq_token_one)
+      create(:candidate, name_first: 'John', name_last: uniq_token_two)
     end
 
-    describe '.create' do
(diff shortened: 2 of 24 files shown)


spec/unit/block_score/api_connection_error_spec.rb
```diff
@@ -1,12 +1,10 @@
...
-  RSpec.describe APIConnectionError do
-    let(:stubbed_route_pattern) { %r{.*api\.blockscore\.com/people/abc123} }
-    let(:stubbed_error) { Errno::ECONNREFUSED }
-    let(:message) { 'Connection refused - Exception from WebMock' }
-    subject { -> { BlockScore::Person.retrieve('abc123') } }
+RSpec.describe BlockScore::APIConnectionError do
+  let(:stubbed_route_pattern) { %r{.*api\.blockscore\.com/people/abc123} }
+  let(:stubbed_error) { Errno::ECONNREFUSED }
+  let(:message) { 'Connection refused - Exception from WebMock' }
+  subject { -> { BlockScore::Person.retrieve('abc123') } }
```
dkubb: The `subject` is evaluated lazily, so there is probably no need to return a `Proc` object here. I realize the shared context would need to be changed but there's only 3 uses that would need to change too.

spec/unit/block_score/candidate_spec.rb
```diff
@@ -1,137 +1,121 @@
...
+  #     and checked. This has the side effect of being susceptible to race
+  #     conditions if tested against the live API in parallel.
+  describe '.all' do
+    let(:uniq_token_one) { '05816347d2234ef4a92465e9784c7ce1' }
+    let(:uniq_token_two) { 'b4cd5fb645788c3b37c11407b83d7741' }
+    subject(:candidates) { described_class.all }
+
+    before do
+      create(:candidate, name_first: 'John', name_last: uniq_token_one)
+      create(:candidate, name_first: 'John', name_last: uniq_token_two)
```
dkubb: I think creating objects would be better done in `let!` blocks rather than a `before` block. Their purpose is to create specific objects to support the test and should be preferred over the more general `before` blocks.

spec/unit/block_score/candidate_spec.rb
```diff
@@ -1,137 +1,121 @@
...
+    subject(:candidate) { create(:watched_candidate) }
+    before do
+      candidate.search
+
+      # ==================================================
+      # Micro Patch: Spinlock until delay job has processed.
+      #
+      # After a new recording, the empty responses should be edited out to
+      #   prevent baking in the sleeps, as they will be replayed in order.
+      VCRHelper.spinlock_until { candidate.hits.count != 0 }
```
dkubb: This inner block would be better written as `candidate.hits.any?`.

spec/unit/block_score/company_spec.rb
```diff
@@ -1,80 +1,60 @@
...
-      end
+  describe '.all' do
+    let(:uniq_token_one) { '4580ab610751f6f37078329357ff98db' }
+    let(:uniq_token_two) { '0f222bcb487f9de689c5c60b854973bc' }
+    subject(:companies) { described_class.all }
 
-      its(:entity_name) { is_expected.to eq 'BlockScore' }
+    before do
+      create(:company, entity_name: uniq_token_one)
+      create(:company, entity_name: uniq_token_two)
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/68/files#r52838300

spec/unit/block_score/invalid_request_error_spec.rb
```diff
@@ -1,10 +1,9 @@
-module BlockScore
-  RSpec.describe InvalidRequestError do
-    subject { -> { BlockScore::Person.create(first_name: 'John') } }
-    let(:message) do
-      '(Type: invalid_request_error) One or more required parameters are invalid (name_first) (Status: 400)'
-    end
-
-    it_behaves_like 'an error'
+RSpec.describe BlockScore::InvalidRequestError do
+  subject { -> { BlockScore::Person.create(first_name: 'John') } }
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/68/files#r52838287

spec/unit/block_score/not_found_error_spec.rb
```diff
@@ -1,10 +1,9 @@
-module BlockScore
-  RSpec.describe NotFoundError do
-    subject { -> { BlockScore::Person.retrieve('abc123') } }
-    let(:message) do
-      '(Type: invalid_request_error) Person with id abc123 could not be found (Status: 404)'
-    end
-
-    it_behaves_like 'an error'
+RSpec.describe BlockScore::NotFoundError do
+  subject { -> { BlockScore::Person.retrieve('abc123') } }
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/68/files#r52838287

spec/unit/block_score/util_spec.rb
```diff
@@ -1,113 +1,111 @@
...
+    end
 
-        it_behaves_like 'parse_json'
+    context 'translates errors' do
+      let(:error) do
+        'An error has occurred. ' \
+          'If this problem persists, please message support@blockscore.com.'
       end
+      let(:input) { '{{' }
+      subject(:parse_attempt) { -> { described_class.parse_json(input) } }
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/68/files#r52838287 (although it's not a shared spec, the same suggestion applies)

spec/unit/block_score/person_spec.rb
```diff
@@ -1,135 +1,114 @@
...
+    let(:uniq_token_two) { 'b19a680671c6c006c6105aa450837668' }
+    subject(:people) { described_class.all }
 
-        it { is_expected.to be_persisted }
-        it { is_expected.not_to be_valid }
-        its(:class) { should be BlockScore::Person }
-      end
+    before do
+      create(:person, name_first: 'John', name_last: uniq_token_one)
+      create(:person, name_first: 'John', name_last: uniq_token_two)
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/68/files#r52838300

spec/unit/block_score/person_spec.rb
```diff
@@ -1,135 +1,114 @@
...
+    subject(:person) { create(:person) }
+    it { expect { person.delete }.to raise_error NoMethodError }
+  end
+
+  describe '#save' do
+    subject(:person) { build(:person) }
+    before { person.save }
+
+    it { is_expected.to be_persisted }
+    it { is_expected.to be_an_instance_of described_class }
```
dkubb: I'm curious what the purpose is behind this kind of test? I mean, doesn't the factory explicitly return this specific type of class? Doesn't that make this a test more about the factory than the `#save` method? Even if we wanted to, in ruby the `#save` method can't change an instance type so this doesn't seem necessary.

However, I don't see any spec for the method's return value, which I think is an important detail and mutant most certainly will flag the lack of return value tests.

spec/unit/block_score/util_spec.rb
tjchambers: @dkubb changing these, but what would be a reason to do it the way it was done?

tjchambers:
@dkubb ready for another pass 

Please note that some of these spec changes will be moved to other specs (they are redundant and are not associated with the units they are testing)

spec/unit/block_score/util_spec.rb
dkubb: > what would be a reason to do it the way it was done?

I'm not sure exactly. It's possible whomever wrote this assumed `subject` was eagerly evaluated. It's a reasonable assumption to make given how many things in RSpec use `subject` implicitly.

PR #71, opened 2016-02-13
Additional Actions specification

tjchambers:


spec/spec_helper.rb
```diff
@@ -45,5 +46,11 @@ module BlockScore
 
     config.order = :random
     Kernel.srand config.seed
+
+    config.around do |example|
+      Timeout.timeout(10) do
```
dkubb: The `example#run` method is also aliased to `#call`, so you could change this to:

```ruby
Timeout.timeout(10, &example)
```

lib/blockscore/collection.rb
```diff
@@ -137,15 +134,17 @@ def retrieve(id)
...
       end
 
-      instance = member_class.retrieve(id)
+      add_instance(member_class.retrieve(id))
+    end
+
+    private
 
+    def add_instance(instance)
       new_member(instance) do |member|
```
dkubb: If you wanted to, you could write this as:

```ruby
def add_instance(instance)
  new_member(instance, &method(:register_to_parent))
end
```

spec/unit/block_score/base_spec.rb
```diff
@@ -14,22 +14,90 @@
       subject(:candidate) { create(:candidate, name_first: 'John') }
       before do
         candidate.name_first = 'Jane'
-        candidate.save
+        result = candidate.save
+        expect(result).to be true
```
dkubb: I don't think there's any reason to assign this to the `result` lvar. This is a [synthetic variable](http://www.perl.com/pub/2000/11/repair3x.html#synthetic variables) and not necessary to solve the problem.

spec/unit/block_score/base_spec.rb
```diff
@@ -14,22 +14,90 @@
...
+      end
+    end
+  end
+
+  describe '#save!' do
+    subject(:candidate) { create(:candidate, name_first: 'John') }
+    context 'when updating an existing candidate' do
+      it 'successfully saves an alteration' do
+        candidate.name_first = 'Jane'
+        result = candidate.save!
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/71/files#r52862517

spec/unit/block_score/base_spec.rb
```diff
@@ -14,22 +14,90 @@
...
+
+        expect(result).to be true
+        expect(candidate.persisted?).to be true
```
dkubb: A slightly nicer way to write this block of code might be:

```ruby
expect { candidate.save! }.to change(candidate, :persisted?)
  .from(false)
  .to(true)
```

The nice thing about that is that it more clearly demonstrates the primary change in the object's `#persisted?` attribute.

spec/unit/block_score/collection_spec.rb
```diff
@@ -26,20 +26,33 @@
...
     let(:question_set_id) { create(:question_set, person_id: person.id).id }
-
-    it { expect(question_sets.retrieve(question_set_id).class).to be BlockScore::Collection::Member }
+    let(:question_set_id2) { create(:question_set, person_id: person.id).id }
+
+    it 'is the proper class' do
+      expect(question_sets.retrieve(question_set_id))
+        .to be_an_instance_of BlockScore::Collection::Member
+      expect(question_sets.retrieve(question_set_id))
+        .to be_an_instance_of BlockScore::Collection::Member
```
dkubb: Did you mean for these 2 lines to duplicate the lines above? I suspect you meant to write this instead:

```ruby
expect(question_sets.retrieve(question_set_id).id)
  .to eql question_set_id.dup
```

spec/unit/block_score/collection_spec.rb
```diff
@@ -26,20 +26,33 @@
...
+      expect(question_sets.count).to equal(1)
```
dkubb: For consistency I would write `.to be(1)`

spec/unit/block_score/collection_spec.rb
```diff
@@ -26,20 +26,33 @@
...
+
+      expect(question_sets.retrieve(question_set_id2))
+        .to be_an_instance_of BlockScore::Collection::Member
+      expect(question_sets.retrieve(question_set_id2).id)
+        .to eql question_set_id2.dup
+      expect(question_sets.count).to equal(2)
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/71/files#r52862894

spec/unit/block_score/person_spec.rb
```diff
@@ -29,22 +31,23 @@
...
 
-      it { expect { described_class.retrieve(person_id) }.to raise_error(ArgumentError, 'ID must be supplied') }
-    end
-
-    context 'empty id presented' do
-      let(:person_id) { '' }
-
-      it { expect { described_class.retrieve(person_id) }.to raise_error(ArgumentError, 'ID must be supplied') }
+    context 'no id presented' do
+      it 'has to have some ID' do
```
dkubb: I think "has to have" is a bit clumsy, what about: `it 'must have a valid id'` ?

spec/unit/block_score/person_spec.rb
```diff
@@ -29,22 +31,23 @@
...
+        expect { described_class.retrieve('') }
+          .to raise_error(ArgumentError, 'ID must be supplied')
+      end
     end
 
     context 'Malformed ID presented' do
       let(:person_id) { 'ABC&&234' }
 
-      it { expect { described_class.retrieve(person_id) }.to raise_error(ArgumentError, 'ID is malformed') }
+      it 'has to have a properly formed ID' do
```
dkubb: What about `it 'must have a valid id'` ?

dkubb:
@tj I finished a pass over this PR.

lib/blockscore/collection.rb
tjchambers: Funny - I was going to ask about that construction.

spec/unit/block_score/collection_spec.rb
tjchambers: This proves a bug was fixed. Previously if the same question_set was retrieved twice it a new question set id would be saved and then a new question set (duplicate content) would be added to the internal list.

spec/unit/block_score/person_spec.rb
tjchambers: We are only really testing to a limited degree it has the right characters - the backend API will tell us if it is truly `valid`? I am concerned about the mixup on the term `valid` because persons are validated by the API. So reading the tests could be confusing.

lib/blockscore/person.rb
```diff
@@ -10,7 +10,7 @@ class Person < Base
 
     attr_reader :question_sets
 
-    def initialize(options = {})
+    def initialize(*)
```
dkubb: I think a more constrained method signature would be `_`. It would assert you can receive "one of anything", rather than "zero or more of anything".

tjchambers: Does that still pass what was received to super?

dkubb: Actually, that's a good point. In this case, `*` and `_` are functionally equivalent due to `super`.

tjchambers: Not that super should be the same, but that's another story.

spec/unit/block_score/actions/delete_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'faker'
+
+RSpec.describe BlockScore::Actions::Delete do
+  describe '#delete' do
+    subject(:candidate) { create(:candidate) }
+    before { candidate.delete }
+
+    it { is_expected.not_to be_persisted }
+    it { expect(candidate.deleted).to be true }
```
dkubb: This isn't something for this PR, but I wonder if the method should be `#deleted?` rather than `#deleted`? The boolean method could be added, and then `#deleted` could use it and/or optionally warn that it is a deprecated method.

lib/blockscore/person.rb
tjchambers: While we are on the topic of super, some people subscribe to calling super last, some first, and some call it when appropriate in the constructor. Which camp are you in?

spec/unit/block_score/actions/delete_spec.rb
tjchambers: There is already a deleted? method. It checks the deleted flag. However the deleted flag (referenced here) is the public API.

lib/blockscore/person.rb
dkubb: I usually call `super` early on. I believe if you're adhering to [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle), the subclass will generally have compatible instance state with it's ancestor class. If there are differences or additions I will perform them after `super`.

It *does* depend on the method though. I follow the above advice for `#initialize` and `.new`. Some methods may require you to step in before `super`, but I find those relatively rare.

tjchambers: Thanks - I believe that is my practice as well.

spec/unit/block_score/actions/delete_spec.rb
dkubb: Oh I know it's an attribute method. I was just saying that even though it's server state I would generally remap it onto an idiomatic method name rather than using the default accessor.

tjchambers: As would I and I have done (need to check to see if predicate is public). But the scary part is it's not immutable.

PR #73, opened 2016-02-16
Update devtools to 1.5.0

tjchambers:
This reverses the constraint commit earlier now that devtools has been updated.


lib/blockscore/response.rb
```diff
@@ -16,13 +16,13 @@ def handle_response(resource, response)
     def api_error(response)
       case response.code
       when 400
-        fail InvalidRequestError.new(response)
+        raise InvalidRequestError.new(response)
```
dkubb: Instead of duplicating the same code 4 times I would probably write this as:

```ruby
def api_error(response)
  error_class = case response.code
  when 400 then InvalidRequestError
  when 401 then AuthenticationError
  when 404 then NotFoundError
  else
    APIError
  end
  raise error_class, response
end
```

Note that if you use `raise` with 2 parameters, you don't have to construct an object.

Also, given that this case statement is a simple lookup table, this could be refactored further into a Hash, eg:

```ruby
HTTP_ERRORS = Hash.new(APIError).update(
  401 => AuthenticationError,
  404 => NotFoundError,
)

# ....

def api_error(response)
  HTTP_ERRORS[response.code]
end
```

tjchambers: Can I take that as an action item ?

tjchambers: I don't understand the #update method usage.

mbj: Maybe you also want to adopt the `fail` convention when raising "new" exceptions.

backus: Yeah we still do prefer `fail` I'm not sure why rubocop changed the default. @dkubb or @tjchambers do you see any value in only using `raise`?

backus:
@tjchambers `raise` comment is my only flag I think

lib/blockscore/actions/retrieve.rb
```diff
@@ -11,7 +11,7 @@ module Actions
     # => #<BlockScore::Person:0x007fe39c424410>
     module Retrieve
       module ClassMethods
-        RESOURCE_ID_FORMAT = /\A[a-f0-9]+\z/.freeze
+        RESOURCE_ID_FORMAT = /\A[a-f0-9]+\z/
```
mbj: I think you still want to freeze regexps, as without `.freeze` the eigenclass could be mutated.

tjchambers: Someone should tell rubocop. It considers it already immutable.

mbj: I think we should not trust rubocop on such issues.

PR #74, opened 2016-02-16
Upgrade Base functions

tjchambers:
Enhance base specs and functionality to support dependent modules

### DIFF
--- a/lib/blockscore/base.rb
+++ b/lib/blockscore/base.rb
@@ -33,7 +33,7 @@ module BlockScore
 
     def save
       save!
-    rescue BlockScore::Error
+    rescue Error
       false
     end
 
--- a/lib/blockscore/base.rb
+++ b/lib/blockscore/base.rb
@@ -38,7 +38,7 @@ module BlockScore
     end
 
     def save!
-      fail BlockScore::Error, "#{resource} is already deleted" if deleted?
+      assert_not_deleted
       capture_attributes(post(endpoint, attributes))
       true
     end
@@ -60,7 +60,15 @@ module BlockScore
       !id.nil? && !deleted?
     end
 
+    def deleted?
+      attributes.fetch(:deleted, false)
+    end
+
     private
+
+    def assert_not_deleted
+      fail Error, "#{resource} is already deleted" if deleted?
+    end
 
     def capture_attributes(source)
       @attributes = source.attributes
@@ -80,10 +88,6 @@ module BlockScore
           attributes[symbol.to_s.chop.to_sym] = value
         end
       end
-    end
-
-    def deleted?
-      attributes.fetch(:deleted, false)
     end
 
     def force!
--- a/spec/unit/block_score/base_spec.rb
+++ b/spec/unit/block_score/base_spec.rb
@@ -31,19 +31,34 @@ RSpec.describe BlockScore::Base, vcr: true do
     end
   end
 
-  describe '#save!' do
+  describe '#deleted?' do
     subject(:candidate) { create(:candidate, name_first: 'John') }
-    context 'when updating an existing candidate' do
+    it 'reflects deleted?' do
+      expect { candidate.delete }.to change(candidate, :deleted?)
+        .from(false)
+        .to(true)
+    end
+  end
+
+  describe '#save!' do
+    subject(:candidate) { build(:candidate, name_first: 'John') }
+    context 'when updating an new candidate' do
       it 'successfully saves an alteration' do
         candidate.name_first = 'Jane'
 
-        expect { candidate.save! } .not_to change(candidate, :id)
+        expect(candidate)
+          .to receive(:post)
+          .with(kind_of(Pathname),
+                hash_including(name_first: 'Jane'))
+          .and_call_original
+        expect(candidate.save!).to be true
         expect(candidate.persisted?).to be true
         expect(candidate.name_first).to eql 'Jane'
       end
     end
 
     context 'when encountering a deleted item' do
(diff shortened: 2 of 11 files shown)


lib/blockscore/base.rb
```diff
@@ -60,8 +60,16 @@ def persisted?
       !id.nil? && !deleted?
     end
 
+    def deleted?
+      attributes.fetch(:deleted, false)
+    end
+
     private
 
+    def already_deleted
```
dkubb: I wonder if a better name for this method would be something like `assert_not_deleted` or `assert_persisted` (although I think the former is probably better)

tjchambers: I didn't like the name either. Of the two I am probably more adept at caching :). I will go with your first choice.

lib/blockscore/base.rb
```diff
@@ -60,8 +60,16 @@ def persisted?
       !id.nil? && !deleted?
     end
 
+    def deleted?
```
dkubb: I don't see any specs for this method. Are you running mutant against the changes you've been making? I would think it would automatically flag this for you.

dkubb: I think we should consider setting up mutant to run as part of the CI process. We could configure it to run on each commit, and test all changes to code made since that commit. This should be all we need to do:

```diff
diff --git a/config/mutant.yml b/config/mutant.yml
index 9a18437..5675a65 100644
--- a/config/mutant.yml
+++ b/config/mutant.yml
@@ -3,3 +3,4 @@ name: blockscore
 namespace: BlockScore
 expect_coverage: 1/1
 ignore_subjects: []
+since: HEAD~1
```

tjchambers: I don't think this is sufficient. That is an inference on my part, but running it without the VCR_RECORD_MODE produces misinformation. My inference is this is not taken into account in the diff above. I do support running mutant, but only if it generates something actionable.

tjchambers: @dkubb apologies - forgot to address your question of `deleted?` coverage. The answer is that I have been looking at the mutation of Base, but it is very slow and large. There are 672 mutations. Since most is private I need to get a gameplan for it together.

dkubb: @tjchambers I enabled mutant to run against each commit and the majority of the things it highlights are missing specs for return values.

Keep in mind the change I made only flags missing mutations in *code changed in the last commit*. It doesn't do anything about existing code -- I suspect we'll need a different plan of attack to resolve missing mutations from existing code. This is more about drawing a line in the sand and not allowing mutation coverage to drop at all in new code.

dkubb:
@tjchambers mutant found some commits in this PR where code was changed without being fully covered by specs. Some of these overlap with #75.

tjchambers:
@dkubb I am going to do the minimum I can for now to make this mutation covered and pass CI. It is a precursor to PR #75. So I am going to work towards fixing the red on this.

tjchambers:
@dkubb this will have to wait till VCR is able to be turned off - it too runs fine without VCR.

dkubb:
@tjchambers It looks like the specs in one of the commits still fails. I assume you have to do a bit more to turn off vcr when mutant is running to make this spec pass.

lib/blockscore/base.rb
```diff
@@ -33,12 +33,12 @@ def inspect
...
       save!
-    rescue BlockScore::Error
+    rescue Error
       false
     end
 
     def save!
-      fail BlockScore::Error, "#{resource} is already deleted" if deleted?
+      assert_not_deleted
       capture_attributes(post(endpoint, attributes))
```
dkubb: There are no specs to make sure the proper attributes are passed to the service.

tjchambers: I am not sure what you mean by this - could you explain further?

dkubb: > I am not sure what you mean by this - could you explain further?

It's demonstrated by the failures mutant is pointing out. Right now mutant can change:

```ruby
capture_attributes(post(endpoint, attributes))
```

To:

```ruby
capture_attributes(post(endpoint, nil))
```

.. without causing a spec failure.

This is *not* a flaw in mutant. Mutant is demonstrating a flaw in the specs. The specs depend on external state, which by definition makes them not unit tests but rather integration tests. I would be more likely to create a `specs/integration` folder and move any vcr (or external service) dependent specs into them and then create a mock connection object for true unit tests.

tjchambers: Thanks for that clarification. I didn't see it as a flaw in mutant. Perhaps the best approach is to move ALL the unit tests to specs/integration and then selectively move/write unit tests for these without VCR. My struggle here is there is a limited contract between the external service and the API and the API and the client software. So for instance a post without content is legitimate - not saying nil is - but certainly posting empty attributes are. 

I have not looked at this mutation coverage as of yet. I did not expect it to be without another pass.

tjchambers:
@dkubb this is ready for another pass

spec/unit/block_score/base_spec.rb
```diff
@@ -31,19 +31,35 @@
     end
   end
 
-  describe '#save!' do
+  describe '#deleted?' do
     subject(:candidate) { create(:candidate, name_first: 'John') }
-    context 'when updating an existing candidate' do
+    it 'reflects deleted?' do
+      expect(candidate.deleted?).to be false
```
dkubb: This could be represented as:

```ruby
expect { candidate.delete }.to change(candidate, :deleted?)
  .from(false)
  .to(true)
```

dkubb:
@tjchambers Sorry for dragging this out. I left one comment.

PR #75, opened 2016-02-16
Enhance deletes specifications

tjchambers:
Add more tests to confirm deeper behavior on delete actions, This includes testing for prior deletions.

PR #76, opened 2016-02-16
Enhance candidate specs

tjchambers:
- Enhance the persisted? tests to affirm true/false responses
- Add a specific save test for iterative saves (candidates are the only type that can be updated over time)
- Make factory specific about the use of watched PERSON
- Extensive additions to the search tests (limited by the return values - still evaluating return values process for test data)

### DIFF
--- a/spec/cassettes/block_score/candidate/create/persisted.yml
+++ b/spec/cassettes/block_score/candidate/create/persisted.yml
@@ -1,0 +1,58 @@
+---
+http_interactions:
+- request:
+    method: post
+    uri: https://BLOCKSCORE_TEST_KEY:@api.blockscore.com/candidates
+    body:
+      encoding: UTF-8
+      string: "{}"
+    headers:
+      Accept:
+      - application/vnd.blockscore+json;version=4
+      User-Agent:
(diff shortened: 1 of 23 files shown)


tjchambers:
@dkubb for your review

dkubb:
@tjchambers there are two commits in this PR that have the subject "Fix delete to have no passed parameters". Was this a mistake in naming, or a mistake with git? Should they be squashed together? If so, should they be part of #75 instead? They seem more related to that PR's changes than this PR.

One other question; does this *have* to include commits from #75? I can understand coupling when you have changes that depend on an earlier PR's changes, but I'm not sure if that's the case here. IMHO if the specs still pass, then I wouldn't bother including other PR's commits.

tjchambers:
@dkubb I have rebased this PR onto master and corrected the duplication. Ready for another pass.

dkubb:
@tjchambers I think you could remove the `Refactor connection authentication` commit since it's in #77.

tjchambers:
@backus @dkubb See my slack message about this CI run. I do not trust the mutant run either, as I cannot see the command being run but infer it is incorrect.

tjchambers:
@dkubb ready for pass - I think this is mergeable now that the commits are corrected and it has been vetted by mutant. Would like this merged.

PR #77, opened 2016-02-16
Refactor connection authentication

tjchambers:
Simplify the connection authentication provisioning and confirmation of API key presence.

### DIFF
--- a/lib/blockscore/connection.rb
+++ b/lib/blockscore/connection.rb
@@ -13,6 +13,12 @@ module BlockScore
       'Content-Type' => CONTENT_TYPE
     }.freeze
 
+    def delete(path, _)
+      request(:delete, path, nil)
+    end
+
+    private
+
     def get(path, params)
       request(:get, encode_path_params(path, params), nil)
     end
@@ -24,12 +30,6 @@ module BlockScore
     def patch(path, params)
       request(:patch, path, params.to_json)
     end
-
-    def delete(path, _)
-      request(:delete, path, nil)
-    end
-
-    private
 
     def request(method, final_endpoint, params)
       begin
--- a/lib/blockscore/connection.rb
+++ b/lib/blockscore/connection.rb
@@ -7,60 +7,51 @@ module BlockScore
     USER_AGENT    = "blockscore-ruby/#{BlockScore::VERSION} (#{REPO})".freeze
     CONTENT_TYPE  = 'application/json'.freeze
     API_URL       = URI.parse('https://api.blockscore.com').freeze
+    HEADERS       = {
+      'Accept'       => ACCEPT_HEADER,
+      'User-Agent'   => USER_AGENT,
+      'Content-Type' => CONTENT_TYPE
+    }.freeze
 
     def get(path, params)
-      request(:get, path, params)
+      request(:get, encode_path_params(path, params), nil)
     end
 
     def post(path, params)
-      request(:post, path, params)
+      request(:post, path, params.to_json)
     end
 
     def patch(path, params)
-      request(:patch, path, params)
+      request(:patch, path, params.to_json)
     end
 
-    def delete(path, params)
-      request(:delete, path, params)
+    def delete(path, _)
+      request(:delete, path, nil)
     end
 
     private
 
-    def headers
-      @headers ||= {
-        'Accept'       => ACCEPT_HEADER,
-        'User-Agent'   => USER_AGENT,
-        'Content-Type' => CONTENT_TYPE
-      }
-    end
-
-    def request(method, endpoint, params)
-      fail NoAPIKeyError, 'No API key was provided.' unless BlockScore.api_key
-
-      path = API_URL + endpoint.to_s
-
+    def request(method, final_endpoint, params)
       begin
-        response = execute_request(method, path, params)
+        response = execute_request(method,
+                                   API_URL + final_endpoint.to_s,
+                                   params)
       rescue SocketError, Errno::ECONNREFUSED => e
-        fail APIConnectionError, e.message
+        fail APIConnectionError, e
       end
 
       Response.handle_response(resource, response)
     end
 
     def execute_request(method, path, params)
-      auth = { username: BlockScore.api_key, password: '' }
+      options = { basic_auth: authentication, headers: HEADERS, body: params }
 
-      params = if method.equal?(:get)
-                 path = encode_path_params(path, params)
-                 nil
-               else
-                 params.to_json
-               end
+      HTTParty.public_send(method, path, options)
+    end
 
-      options = { basic_auth: auth, headers: headers, body: params }
-
-      HTTParty.send(method, path, options)
+    def authentication
+      fail NoAPIKeyError, 'No API key was provided.' unless BlockScore.api_key
+      { username: BlockScore.api_key }
     end
 
     def encode_path_params(path, params)
--- a/spec/unit/block_score/actions/delete_spec.rb
+++ b/spec/unit/block_score/actions/delete_spec.rb
@@ -39,4 +39,17 @@ RSpec.describe BlockScore::Actions::Delete, vcr: true do
       end
     end
   end
+
+  describe '#delete!' do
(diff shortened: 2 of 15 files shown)


tjchambers:
@dkubb for your review

spec/unit/block_score/actions/delete_spec.rb
```diff
@@ -3,9 +3,35 @@
...
+    context 'previously deleted' do
+      it 'was already deleted' do
+        expect(candidate.delete).to be false
+      end
+    end
+  end
+
+  describe '#delete!' do
+    subject(:candidate) { create(:candidate) }
+    before do
```
mbj: I'd add a blank line between subject and before to improve readability.

tjchambers: Thanks - I intend to make a style pass later.

dkubb: When you get to that point lmk. I have a few style guidelines wrt spec layout that can make things more consistent, and improve the readability.

dkubb:
@tjchambers I'd recommend removing commits from this that aren't specific to this PR and/or required to make the specs pass. As near as I can tell there's only the last commit in the series, but I wanted to be sure.

tjchambers:
@dkubb I have rebased this on master and removed the other commits. Ready for another pass.

dkubb:
@tjchambers mutant found a commits in this PR where code was changed without being fully covered by specs. This overlaps with #76, although I think you can safely remove this commit from that PR without negatively affecting it's functionality.

tjchambers:
@dkubb This commit fixes mutation coverage but ONLY if not running VCR. So this will need to be put on top of the VCR fix (your earlier patch), rebased and then rerun through CI for confirm.

dkubb:
> So this will need to be put on top of the VCR fix (your earlier patch), rebased and then rerun through CI for confirm.

@tjchambers Do you mean the one I pasted as a snippet in the private chat, the one where I add `vcr: true` to all the `RSpec.describe` blocks?

tjchambers:
@dkubb yes - I mean this should go on that patch, which I see you have a branch for. If we PR that branch then I can begin the task of adjusting VCR usage so that we can appropriately get the rest of these PR's across the mutation coverage hurdle.

tjchambers:
@dkubb ready for review. 

I alerted @john to the fact that it looks like all of the builds are running mutant twice now. When this PR was run with -j4 the first time it failed, and when I just rebuilt it it ran fine.

lib/blockscore/connection.rb
```diff
@@ -1,66 +1,57 @@
 module BlockScore
   module Connection
     MAJOR_VERSION = BlockScore::VERSION.split('.').first
-    REPO          = 'https://github.com/BlockScore/blockscore-ruby'.freeze
-    API_VERSION   = "version=#{MAJOR_VERSION}".freeze
+    REPO = 'https://github.com/BlockScore/blockscore-ruby'.freeze
+    API_VERSION = "version=#{MAJOR_VERSION}".freeze
```
dkubb: As a general rule, when I have constants or local variables on consecutive lines, all using assignment operators, I will align the assignment operators.

lib/blockscore/connection.rb
```diff
@@ -1,66 +1,57 @@
...
-    USER_AGENT    = "blockscore-ruby/#{BlockScore::VERSION} (#{REPO})".freeze
-    CONTENT_TYPE  = 'application/json'.freeze
-    API_URL       = URI.parse('https://api.blockscore.com').freeze
+    USER_AGENT = "blockscore-ruby/#{BlockScore::VERSION} (#{REPO})".freeze
+    CONTENT_TYPE = 'application/json'.freeze
+    API_URL = URI.parse('https://api.blockscore.com').freeze
+    HEADERS = {
+      'Accept' => ACCEPT_HEADER,
+      'User-Agent' => USER_AGENT,
+      'Content-Type' => CONTENT_TYPE
```
dkubb: I would suggest aligning the `=>` symbols for readability.

spec/unit/block_score/connection_spec.rb
```diff
@@ -0,0 +1,97 @@
+require 'faker'
+
+RSpec.describe BlockScore::Connection do
+  before do
+    BlockScore.api_key = ENV.fetch(BlockScore::ENVIRONMENT_API_KEY,
```
dkubb: This is setting a global variable in a before block, but it doesn't reset the variable back to the original value afterwards. Are there other tests that might depend on the value of this global variable?

tjchambers: Yes - they all do. It is reestablished in a before(:context) block universally.

lib/blockscore/connection.rb
tjchambers: I need to sit down and have a talk with my IDE.

lib/blockscore/connection.rb
tjchambers: See above - they *were* lined up at one point.

spec/unit/block_score/connection_spec.rb
dkubb: Ahh ok good, just checking to be sure.

lib/blockscore/connection.rb
tjchambers: BTW is there a rubocop setting for that?

dkubb: I'm not sure. I would hope so, but I can't recall looking for it since it's a habit I've carried over from Perl.

tjchambers: I like it aligned - so I need to research.

tjchambers:
@dkubb ready for another pass.

lib/blockscore/connection.rb
```diff
@@ -7,60 +7,51 @@ module Connection
...
     API_URL       = URI.parse('https://api.blockscore.com').freeze
+    HEADERS       = {
+      'Accept'       => ACCEPT_HEADER,
+      'User-Agent'   => USER_AGENT,
+      'Content-Type' => CONTENT_TYPE
+    }.freeze
 
-    def get(path, params)
-      request(:get, path, params)
+    def delete(path, _)
```
dkubb: Why does the `#delete` method take any parameters at all?

tjchambers: It doesn't this was an interim step to get to the point where I could refactor it and not conflict with all the PR's

dkubb: Or should it accept `params` and then do `encode_path_params(path, params)` like `#get`?

tjchambers: It has no params.

lib/blockscore/connection.rb
```diff
@@ -7,60 +7,51 @@ module Connection
...
-
+    def request(method, final_endpoint, params)
       begin
-        response = execute_request(method, path, params)
+        response = execute_request(method,
+                                   API_URL + final_endpoint.to_s,
+                                   params)
       rescue SocketError, Errno::ECONNREFUSED => e
-        fail APIConnectionError, e.message
+        fail APIConnectionError, e
```
dkubb: This is a good change. I dislike it when people throw an exception, and then throw away the original exception object when re-throwing something else. It makes it really difficult to debug without a proper stacktrace.

tjchambers: Unfortunately the only thing retained is the message.

dkubb: > Unfortunately the only thing retained is the message.

You can retain the backtrace by passing it in as the 3rd argument to raise, eg:

```ruby
fail APIConnectionError, e, e.backtrace
```

Also, I would recommend against single letter variable names, even in this specific instance.

dkubb: I'm not going to hold back merging this PR for this change, but I'd recommend it for the future because it can sometimes be quite difficult to find the cause of an error.

When the time comes, it might almost be superior to use something like `fail APIConnectionError.new(e)`, and then we can have it actually hold onto the actually error object rather than unpacking it and extracting it's message.

PR #79, opened 2016-02-18
Add mutant to CI process

dkubb:
This branch ensures that every commit is tested to make sure it does not introduce a coverage regression.

### DIFF
--- a/circleci.sh
+++ b/circleci.sh
@@ -28,5 +28,5 @@ if [[ "$1" = "build" ]]; then
 elif [[ "$1" = "spec" && "$BUILD" = "tests" ]]; then
   rvm-exec $CIRCLE_NODE_RUBY_VERSION rspec;
 elif [[ "$1" = "spec" && "$BUILD" = "metrics" ]]; then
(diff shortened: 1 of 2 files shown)


backus:
My only comment / question is non-blocking so I will merge now. If you look at container #3 though for the build it looks like there is a bunch of stray warnings from MRI about rubocop's source. I assume this is just due to rubocop being required by default through devtools. To make that build more readable maybe there would be a way to specify `require: false` in the gemfile for rubocop so that these warnings don't pile up in the unrelated mutant build?

backus:
@dkubb actually one more note: I think since @tjchambers is still likely to have PRs which change tests without changing code, you might still need to be diligent in those cases since mutant will not be triggered by spec only changes

dkubb:
> I think since @tjchambers is still likely to have PRs which change tests without changing code, you might still need to be diligent in those cases since mutant will not be triggered by spec only changes

Good point. I will look for cases where specs only were changed but not the code. I'm not *quite* as concerned by this as code changed without specs, but it's something I will be on the lookout for.

PR #80, opened 2016-02-18
Add metrics:mutant rake task that runs mutant with a single job

dkubb:


dkubb:
@tjchambers this is ready for review.

mbj:
Curious why this change is needed? I can see that the mutant default does not work well in travis as the host container CPU cont is much higher than the containers allocated CPU, so mutant spawns way to many threads.

Could you go with a higher number? like `-j 4` or something? Or do you have indeed a cross killfork visible shared mutable resource?

tjchambers:
@dkubb I presume since at present this includes no VCR disablement universally for the CI, that that will be addressed through config and vcr: true/false on individual cases of code, making the disablement of VCR case specific and same both in CI and outside CI. If so I am fine with that. 

The reduction to -j 1 is only necessary under the situation where we are NOT using VCR AFAICT. This sets the stage for VCR disablement.

dkubb:
> I presume since at present this includes no VCR disablement universally for the CI, that that will be addressed through config and vcr: true/false on individual cases of code, making the disablement of VCR case specific and same both in CI and outside CI

Correct. I wasn't sure I could do it in a way that was global and consistent.

> The reduction to -j 1 is only necessary under the situation where we are NOT using VCR AFAICT. This sets the stage for VCR disablement.

Should we delay merging, or would you like this in the project sooner rather than later?

tjchambers:
We can merge this IMO at any time. I would prefer sooner to lower the risk of race conditions interfering in mutant testing.

mbj:
> The reduction to -j 1 is only necessary under the situation where we are NOT using VCR AFAICT. This sets the stage for VCR disablement.

I could see a more fine grained config (and the option to read it from a file) could help here to set the concurrency specific to certain subjects / tests. Certainly not a feature for the free version of mutant. But I see growing need for it.

tjchambers:
Yes - it is more module/example specific than universal. The CLI config disables what otherwise might be quite an opportunity for acceleration most of the time.

PR #81, opened 2016-02-19
Add vcr selector to specs

dkubb:
When this selector is present the test will be wrapped by vcr. In the future the specs will be refactored to not all depend on vcr, and this is the first step towards this ideal.

### DIFF
--- a/spec/unit/block_score/actions/all_spec.rb
+++ b/spec/unit/block_score/actions/all_spec.rb
@@ -1,6 +1,6 @@
 require 'faker'
 
-RSpec.describe BlockScore::Actions::All do
+RSpec.describe BlockScore::Actions::All, vcr: true do
   describe '.all' do
     let(:uniq_token_one) { '4a2c019a1e3c33442644d9f52c7a93f7' }
     let(:uniq_token_two) { 'ba9a680671c6c006c6105aa450837668' }
--- a/spec/unit/block_score/actions/create_spec.rb
+++ b/spec/unit/block_score/actions/create_spec.rb
@@ -1,6 +1,6 @@
 require 'faker'
 
-RSpec.describe BlockScore::Actions::Create do
+RSpec.describe BlockScore::Actions::Create, vcr: true do
   describe '.create' do
     subject(:candidate) do
       BlockScore::Candidate.create(attributes_for(:candidate))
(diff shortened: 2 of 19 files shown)


dkubb:
@tjchambers this is ready for review/merge.

tjchambers:
@dkubb my concern is how has this been tested to confirm we get the behavior we desire? I don't see any confirmation tests (vcr: true, vcr: false, :vcr missing...). I am not asking for any tests (well maybe I am), but am more questioning how do we know the vcr_setup change gets the desired effect.

tjchambers:
@dkubb I would be alright just knowing that there is no regression from putting this in.

tjchambers:
@dkubb I am going to merge this and deal with any potential regressions after rebasing.

dkubb:
> my concern is how has this been tested to confirm we get the behavior we desire? I don't see any confirmation tests (vcr: true, vcr: false, :vcr missing...). I am not asking for any tests (well maybe I am), but am more questioning how do we know the vcr_setup change gets the desired effect.

If you toggle the `vcr: true` in `spec/vcr_helper.rb` to `vcr: false` you'll see how vcr is turned off globally across the project since it won't match anything. You can also turn it off on a per-spec basis by removing it and running the spec. It should fail in the same way.

I usually don't write tests for spec configurations, but I usually make it strict enough that if something doesn't behave as expected an error is thrown. I'm not exactly sure what the best approach would be here though. It would be awesome for vcr to flag cases where a specific cassette wasn't used so that we'd know that if the specs were successful that every spec with `vcr: true` configured used every recorded spec.

tjchambers:
I don't find it working that way.

tjchambers:
@dkubb my inference is that VCR is ON unless turned off. So merely not running all specs through the process is actually worse, because VCR is on but can't find the cassettes.

tjchambers:
@dkubb I think I see how it has to work, so I am going to test it locally.

tjchambers:
@dkubb here is what seems to work properly:

```
    config.around(:each) do |example|
      record_option = ENV.fetch(RECORD_MODE_KEY, :none).to_sym
      vcr_on = example.metadata.fetch(:vcr, false)
      options = { record: (vcr_on ? record_option : :skip) }
      BlockScore.run_vcr_example(options, example)
    end
```

tjchambers:
@dkubb since I need to make another correction to vcr_setup I am going to make a fix branch and include 2 commits, PR for your review.

PR #82, opened 2016-02-20
Correct vcr setup

tjchambers:
Since VCR needs to be explicitly turned off, the prior approach of just running without VCR is insufficient. VCR needs to be "skipped" along with full external network access re-enabled if vcr: true is missing. This commit fixes that, so with vcr true then the recording mode is fetched from environment variable (as was prior). If vcr: true is missing, then the recording mode is set to :skip which triggers the enablement of the network and VCR is explicitly turned off for that example.

### DIFF
--- a/spec/vcr_setup.rb
+++ b/spec/vcr_setup.rb
@@ -9,7 +9,7 @@ module BlockScore
   extend self
 
   RSpec.configure do |config|
-    config.before(:suite) do
+    config.before(:context) do
       BlockScore.api_key = ENV.fetch(ENVIRONMENT_API_KEY, PLACEHOLDER_API_KEY)
     end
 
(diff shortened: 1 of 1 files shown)


tjchambers:
@dkubb when green, for your review

dkubb:
@tjchambers I can understand using this as a stop-gap, but I think it's a mistake (long term) to have unit tests that can access the network at all. Even in an integration test, the worst I could ever accept is perhaps a way of allowing a test to opt-in to being able to access a specific server, and then only for the purpose of recording a cassette.

The problems we're seeing with the specs and mutant is mostly due to the fact that we have integration tests that are standing-in for unit tests. A true unit test wouldn't be dependant on an external service to specify it's correctness.

Long term what I would like to see is a way to swap out the connection object with something that can implement some basic checks, and has a way to introspect it's state after some method call to make sure it received the expected messages. The normal connection object could still be integration tested using vcr, but it's unit tests would use shared examples that the mock connection object also passes.

PR #83, opened 2016-02-20
Upgrade action specs

tjchambers:
Also fix to move immutable attribute processing into Base.


tjchambers:
@dkubb ready for review

mbj:
I really wounder if its possible to reduce the churn, the diffstat of: `+23,049 −300` makes me think about something must have gone wrong, or be wrong with this project.

I realize most of these comes from generated VCR captures, but if you only do a ~50loc change to tests how could you end up so many VCR requests being re-generated?

Maybe old requests need to be pruned so the additions also come with deletions. Also maybe `ffaker` can be tuned to return deterministic results till you factored out VCR?

tjchambers:
@mbj most of the churn from what I see is the changes in the RSpec.describe, which invalidated many of the original VCR cassettes because the cassette names changed. The only positive from VCR is that the specs run fast. I won't list the dozens of negatives and workarounds. In many cases the newer specs also have whitespace issues, which means that the falsley report a change in the whitespacing.

mbj:
> @mbj most of the churn from what I see is the changes in the RSpec.describe, which invalidated many of the original VCR cassettes because the cassette names changed. The only positive from VCR is that the specs run fast. I won't list the dozens of negatives and workarounds. In many cases the newer specs also have whitespace issues, which means that the falsley report a change in the whitespacing.

Maybe the now unused recordings should be deleted? So its not 24kloc additions alone, and maybe paired with 24kloc deletions?

tjchambers:
@mbj the old ones should be deleted. But with so many moving parts, I am trying to avoid considering the conscious deleting of unused/replaced VCR cassettes for each commit. A slight change in the describe text creates a fresh cassette and and unused old one (perhaps). I feel sometimes like all I am doing is working around the process instead of making substantive tests and real code changes.

backus:
> @mbj the old ones should be deleted. But with so many moving parts, I am trying to avoid considering the conscious deleting of unused/replaced VCR cassettes for each commit. A slight change in the describe text creates a fresh cassette and and unused old one (perhaps). I feel sometimes like all I am doing is working around the process instead of making substantive tests and real code changes.

I've mentioned this before but deriving cassette names from the `RSpec.describe` statement was a choice the previous author made which you can certainly change if you are finding yourself working around it.

dkubb:
@tjchambers I attempted to rebase this on top of master but I saw spec failures, so I left it as-is so you could do it.

tjchambers:
thanks - I will take care of it

lib/blockscore/actions/update.rb
```diff
@@ -15,32 +15,17 @@ module Actions
...
+        persisted? ? update_it : super()
       end
 
-      # Filters out the non-updateable params.
-      def filter_params
-        # Cannot %i syntax, not introduced until Ruby 2.0.0
-        attributes.reject { |key, _| PERSISTENT_ATTRIBUTES.include?(key) }
+      private
+
+      def update_it
```
dkubb: Why not just use the name `#update` ?

tjchambers: My recall is it confused mutant because `update` is a method name farther up the chain.

lib/blockscore/base.rb
```diff
@@ -66,18 +67,20 @@ def capture_attributes(source)
...
         define_method(symbol) do
-          wrap_attribute(attributes[symbol])
+          wrap_attribute(attributes.fetch(symbol))
         end
       end
     end
 
-    def add_setter(symbol, *_args)
+    def add_setter(symbol)
+      attr = symbol.to_s.chop.to_sym
```
dkubb: Just to confirm, the `String#chop` is removing the trailing `=` from the method name? I would be more inclined to use `symbol.to_s.chomp('=').to_sym` so it's more clear what is actually getting removed. The `'='` could even be moved into a frozen constant.

tjchambers: You are correct - that is the purpose. That would be a good approach. Will change.

lib/blockscore/base.rb
```diff
@@ -66,18 +67,20 @@ def capture_attributes(source)
...
+      assert_mutable(attr)
       singleton_class.instance_eval do
```
dkubb: Not in this PR, but I definitely think we should examine this block. By modifying the singleton like this it is going to blow the global method cache. This affects everything in the whole runtime, not just this specific code.

A better approach would be to define the attributes up-front per-class, and then create the methods when the class is initialized.

spec/unit/block_score/actions/all_spec.rb
```diff
@@ -11,5 +11,57 @@
...
+    it 'extends the klass' do
+      expect(klass.singleton_class)
+        .not_to include(described_class::ClassMethods)
+      klass.send(:include, subject)
+      expect(klass.singleton_class)
+        .to include(described_class::ClassMethods)
+    end
+  end
+
+  describe '.included' do
```
dkubb: Did you mean to duplicate this block?

spec/unit/block_score/actions/create_spec.rb
```diff
@@ -9,4 +9,18 @@
     it { is_expected.to be_persisted }
     it { is_expected.to be_an_instance_of BlockScore::Candidate }
   end
+
+  describe '.included' do
```
dkubb: You could probably move this into a shared example and reuse it in the affected specs.

spec/unit/block_score/actions/retrieve_spec.rb
```diff
@@ -25,4 +25,18 @@
       end
     end
   end
+
+  describe '.included' do
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/83/files#r53864537

spec/unit/block_score/actions/all_spec.rb
```diff
@@ -11,5 +11,57 @@
...
+        '(Status: 400)'
+      end
+      it 'returns an error if comparison operator not in allowed list' do
+        expect { BlockScore::Person.all('created_at[gtx]' => 1_455_660_769) }
+          .to raise_error(BlockScore::InvalidRequestError, expectation)
+      end
+    end
+  end
+
+  describe '.included' do
```
dkubb: Instance of https://github.com/BlockScore/blockscore-ruby/pull/83/files#r53864537

lib/blockscore/base.rb
tjchambers: What you are proposing sounds like you would have to know what attributes are permitted.

spec/unit/block_score/actions/all_spec.rb
tjchambers: No - it sounds like it may be an artifact perhaps of rebasing? Just guessing.

spec/unit/block_score/actions/update_spec.rb
```diff
@@ -2,15 +2,41 @@
...
+          end)
       end
+    end
 
-      it { is_expected.to be_persisted }
-      its(:name_first) { is_expected.to eql 'Jane' }
+    context 'when changing first update candidate', vcr: { record: :all } do
+      it 'has the same attributes' do
+        expect(candidate.persisted?).to be false
+        expect(candidate.save!).to be true
```
dkubb: I would use `expect { candidate.save! }.to change(candidate, :persisted?).from(false).to(true)` here since this part is testing the behaviour of the `#save!` method and how it changes the `#persisted?` state.

tjchambers: Yes - that is an idiom I need to use more often.

spec/unit/block_score/actions/all_spec.rb
dkubb: I assumed so, just wanted to bring it to your attention.

dkubb: Usually this kind of thing happens when there is a merge conflict that is resolved manually. I don't think it could happen in the course of a normal rebase where there is no conflict.

lib/blockscore/base.rb
dkubb: > What you are proposing sounds like you would have to know what attributes are permitted.

Do we not know what attributes are permitted in the API?

tjchambers: Yes and no. Presently it is designed such that anything that is a valid method name can be used as an attribute. The backend decides whether it is going to silently discard it. And AFAICT that is desired behavior. Personally I would want to know about typos if I was a developer. We could probably define the extent of the values, but moving that to the API code means it has to be changed everytime the back end changes. Not saying that is often, but they need to obviously stay in sync. So far when I have floated that concept it seems like that is reserved for next version of API.

spec/spec_helper.rb
```diff
@@ -48,7 +48,7 @@ module BlockScore
     Kernel.srand config.seed
 
     config.around do |example|
-      Timeout.timeout(10, &example)
+      Timeout.timeout(6, &example)
```
dkubb: Why adjust this timeout?

OT I should mention that when I was working from a coffeeshop I had to locally edit this in order to allow the specs to pass due to a less than favourable speed connection.

tjchambers: In our conversation the other day we talked about pushing it down to 3 seconds. It is excessive at 6 seconds IMO.

spec/unit/block_score/actions/update_spec.rb
tjchambers: In this particular case I cannot use the idiom, because I need to test the return value to satisfy killing a mutation.

mbj: > In this particular case I cannot use the idiom, because I need to test the return value to satisfy killing a mutation.

As there is the `.to(true)` block the return value is verified, and changing the value to anything but `true` should be killed.

If the mutation is still alive using `expect { }.to change` might use a to permissive internal matching on expected values, that would be really unfortunate.

spec/spec_helper.rb
mbj: The lower this timeout is the better from a mutation testing perspective, as infinite loops are killed earlier.

It comes as a downside that a loop might be killed for "assumed infinite loop" where it would actually terminate with a green test so an alive mutation. So there is a trade off.

spec/unit/block_score/actions/update_spec.rb
tjchambers: The return value that needs testing is from #`save!`and should only be captured once.

mbj: Ahh I see.

What about:

```
expect { expect(candiate.save!).to be(true) }.to change(candidate: persisted?).from(false).to(true).
```

tjchambers: I wasn't expecting that to be good style.

mbj: Its better semantic specification. Trading better semantics for less good style (you can also do a multi line version with `do/end` to make it more readable) is IMO acceptable.

dkubb: I agree this could work, although I do think the specs are a bit coarse grained. In a *true* unit test I would only test one specific behaviour. The return value would be tested separately from the state change behaviour.

Given these are integration tests its ok to test a few things in a specific `it` block given how expensive the tests are to execute. In a real unit test I would want a failure of a given `it` block to be self-evident -- something which is difficult when it tests multiple unrelated aspects of a method.

tjchambers:
@dkubb this is ready for another pass

PR #85, opened 2016-02-25
VCR cassettes refresh

tjchambers:
Clear out old specs from block_score directory and below and regenerate them in the new proper location.

This is based on top of PR #83


tjchambers:
@dkubb - this is a cleanup of the current VCR cassettes. It is based on #83

dkubb:
@tjchambers I added another commit that cleans up all the remaining vcr cassettes.

Since it's just a removal I'll merge the PR once all the specs are green.

mbj:
@dkubb What about adding a check on CI there are no stale VCR cassettes? This would force not to introduce such "dead fixture data regressions".

tjchambers:
@mbj I didn't know that could be done.

